PR #7, opened 2023-02-09
DM-37933: Update Python packaging

rra:
Bring the Python build system configuration and packaging in line with the current `fastapi_safir_app` template, specifically:

- Require Python 3.11 as the minimum version and update the Docker and CI configuration accordingly
- Stop pushing containers to Docker Hub since we use GitHub Container Registry by preference now
- Drop `documentation_url` since we don't currently build a documentation site
- Convert `README.rst` to `README.md`, remove some obsolete information, link to SQR-072, and link to the published version of SQR-066 since we're about to merge the changes to make it match this implementation
- Simplify the tox configuration since only the test target should need the environment variable setting
- Switch to a `pyproject.toml`-only build system and drop `setup.cfg` and `setup.py`
- Remove the explicit `MANIFEST.in` entry for assets since this is only required when they haven't been committed. Once committed, setuptools_scm should handle this automatically.


athornton (approved):
This all looks great.

PR #8, opened 2023-02-09
DM-37933: Remove utility functions that are in Safir

rra:
Safir provides CamelCaseModel and a current_datetime utility function, so use those functions from there instead of defining new ones.

Switch the index metadata to BaseModel, since CamelCaseModel is not serving any purpose in models used only for the web API.

Remove validate_exactly_one_of, which wasn't used anywhere.


athornton (approved):
This one all looks good too.

PR #10, opened 2023-02-09
DM-37933: Better testing of tag parsing

rra:
Import the tag parsing tests from cachemachine. There were almost no tests of the tag parsing logic and this (for now until we create a shared model library) will be the canonical location for this code, so import the tests from cachemachine and reformat them a little to be less verbose. Anticipate eventually moving the tag out of models to its own abstract data type in the naming of the test suite.

Refactor the construction of the build component in tags to fix a bug I noticed (`|` were left behind in the build component) and make the code a bit easier to follow.


athornton (approved):
Amazing how much simpler _build_component got.

PR #11, opened 2023-02-09
DM-37933: Add tests for the Docker storage layer

rra:
Test the code to retrieve a tag list and an image digest from the Docker API using mocked httpx responses.  This introduces a mocking system for the Docker API that can eventually be used to retire the higher-level mock of the Docker storage object.


athornton:
I did not know about respx.  That does indeed make it pretty easy to mock the Docker client.

athornton:
Oh.  Heh.  I came back here after reading some stuff about respx to say "yeah, it all looks good" but you beat me to it.

rra:
Oh, whoops, sorry, I missed that you hadn't approved it yet!

PR #12, opened 2023-02-09
DM-37933: Simplify and further test Docker credential handling

rra:
Rather than using separate credentials for the Docker storage layer test, plumb through the dynamic loading of credentials and use the default credentials for the test configuration, along with the default registry and repository. This enables creating the Docker storage object via the factory, so provide a test factory as a fixture and add a method to create a Docker storage object.

Simplify the DockerCredentials class to stop storing username and password in two different forms and instead derive the other forms from username and password.

This work uncovered a latent minor bug in the Docker storage layer, where a spurious extra slash was added to the URL for getting tag and image information. The Docker API apparently swallows this but respx doesn't, so redo the repository and path properties of the prepuller configuration to remove the unwanted leading slash in the repository.

### DIFF
--- a/src/jupyterlabcontroller/factory.py
+++ b/src/jupyterlabcontroller/factory.py
@@ -128,16 +128,25 @@ class Factory:
         async with aclosing(factory):
             yield factory
 
-    def __init__(
-        self,
-        context: ProcessContext,
-        logger: BoundLogger,
-    ) -> None:
+    def __init__(self, context: ProcessContext, logger: BoundLogger) -> None:
         self._context = context
         self.logger = logger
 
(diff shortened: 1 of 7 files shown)


athornton (approved):
I like this, but I'd also suggest one more enhancement:

Add a classmethod and a regular method to the DockerCredentials class: the classmethod would allow construction of DockerCredentials (or, if I remember the format correctly, a dict of domain-to-docker-credentials) from a .dockerconfigjson file, and the corresponding method to generate the .dockerconfigjson file from the credentials.  This would ease migration and give us more confidence when adding new auth domains, which has always been a fraught process.

PR #13, opened 2023-02-10
DM-37933: Config and Docker credentials refactor

rra:
The runtime section of the `Configuration` object had various parameters that were determined during configuration load and then set after creating the object, which required awkward initialization that was then ignored. All of these parameters could be handled a different way:

- `instance_url` was really `base_url` and needed to be initialized from an environment variable, which allowed it to be a normal top-level configuration parameter as long as the top-level configuration was changed to inherit from `BaseSettings` so that it would check the environment.
- `namespace_prefix` could be initialized from a `default_factory` function and was therefore moved to the lab configuration, since that's where it's used. (It's also reused as the namespace for prepulling, but shouldn't be; this will be fixed later.)
- `path` was only used to decide where to load Docker credentials from, which was better handled by a top-level configuration parameter with a default matching the expected Kubernetes path that is overridden by the test suite.

To enable this change, and improve the storage and service separation, overhaul how Docker secrets are managed. Replace `DockerCredentialsMap` with a new `DockerCredentialStore` class that (for the future) supports setting new credentials and saving the credential store as well. Rework `DockerStorageClient` to support multiple repositories in the same object (which we will need if we eventually support prepulling multiple images), with the registry and repository passed as arguments to its methods. Rework how it handles authentication to avoid the confusing `authenticate` bool parameter to its methods, which was not really part of the external API.

### DIFF
--- a/src/jupyterlabcontroller/services/prepuller/docker.py
+++ b/src/jupyterlabcontroller/services/prepuller/docker.py
@@ -1,26 +1,83 @@
-"""This class will manage communication with the Docker repository in order
-to fetch tag information and determine which images are available from the
-repository.
+"""Docker client for image prepuller."""
 
-These are known as our remote images.
-"""
+import asyncio
 
+from structlog.stdlib import BoundLogger
 
+from ...models.tag import RSPTag, RSPTagList, TagMap
+from ...models.v1.prepuller_config import PrepullerConfiguration
 from ...storage.docker import DockerStorageClient
 from .state import PrepullerState
 
 
 class PrepullerDockerClient:
+    """Query Docker for available images.
+
+    Fetches tag information for a given repository from a Docker registry.
+    These are known as our remote images and are used to calculate which
+    images to prepull.
+    """
+
     def __init__(
         self,
         namespace: str,
         state: PrepullerState,
         docker_client: DockerStorageClient,
+        config: PrepullerConfiguration,
+        logger: BoundLogger,
     ) -> None:
-        self.logger = docker_client.logger
+        self.logger = logger
         self.state = state
         self.namespace = namespace
         self.docker_client = docker_client
+        self.config = config
+
+    async def get_tag_map(self) -> TagMap:
+        """Get a list of all available tags.
+
+        Query all the tags for a given repository to get their digests,
+        inflate the tag and digest pairs into RSPTag objects, and return a
+        structure with dicts indexed by tag and by digest.  This in turn will
+        let us easily compare what we have remotely and what we have locally,
+        because it doesn't matter what tag we pulled a given image by, but
+        whether we have an image with the proper digest already on a given
+        node.
+        """
+        # Get all the tags and digests from the Docker API.  This is much too
+        # aggressive in parallelism and also is doing too much work since we
+        # don't care about the digests for most tags.
+        tags = await self.docker_client.list_tags(
+            self.config.registry, self.config.repository
+        )
+        tasks = [
+            asyncio.create_task(
+                self.docker_client.get_image_digest(
+                    self.config.registry, self.config.repository, tag
+                )
+            )
+            for tag in tags
+        ]
+        digests = await asyncio.gather(*tasks)
+        digest_for_tag = dict(zip(tags, digests))
+
+        # Inflate each tag/digest pair into an RSPTag
+        rsp_tags = []
+        for tag, digest in digest_for_tag.items():
+            alias_tags = []
+            if tag == self.config.recommended_tag:
+                alias_tags = [self.config.recommended_tag]
+            rsp_tags.append(
+                RSPTag.from_tag(
+                    tag=tag,
+                    digest=digest,
+                    image_ref=f"{self.config.path}:{tag}",
+                    alias_tags=alias_tags,
+                )
+            )
+
+        # Put it into a RSPTagList, which will sort the tags and produce
+        # the map.
+        return RSPTagList(tags=rsp_tags).tag_map
 
     async def refresh_if_needed(self) -> None:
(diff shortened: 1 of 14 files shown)


athornton:
Er: shouldn't be used as the namespace for prepulling?  I thought we agreed that prepulled pods definitely belong in the same namespace as the jupyterlab controller.  Which, maybe that's not the right term for it, but "namespace prefix" is where all the pod-spawning stuff lives (controller, hub, and lab), and then that-username is where user pods go, which then lets us do different spawners side by side, which is nothing you want in steady state but is very useful as you're migrating.

rra:
The problem isn't the content, it's the semantics. There are really two separate settings here. One is the prefix to add to the username to form a namespace for the user's lab environment. The other is the namespace into which to put prepulled pods (which should be the same namespace in which the lab controller is running).

These don't conceptually need to be the same thing. For example, someone might want all the user environments in namespaces called `users-<username>`. That does not mean that the prepuller should put its pods in a namespace called `users` (which may not even exist).

The prepuller should just always use the current namespace, independent of the configuration of the namespace prefix.

src/jupyterlabcontroller/services/prepuller/docker.py
```diff
@@ -1,34 +1,91 @@
...
+        inflate the tag and digest pairs into RSPTag objects, and return a
+        structure with dicts indexed by tag and by digest.  This in turn will
+        let us easily compare what we have remotely and what we have locally,
+        because it doesn't matter what tag we pulled a given image by, but
+        whether we have an image with the proper digest already on a given
+        node.
+        """
+        # Get all the tags and digests from the Docker API.  This is much too
+        # aggressive in parallelism and also is doing too much work since we
+        # don't care about the digests for most tags.
```
athornton, line 48: I'm not at all sure we don't.  We will need it to determine which tags are aliases for other tags, and since we have it we can detect whether we have stale images stored locally.

src/jupyterlabcontroller/storage/docker.py
```diff
@@ -1,202 +1,313 @@
...
-        if r.status_code == 200:
-            body = r.json()
-            return body["tags"]
-        elif r.status_code == 401 and authenticate:
-            await self._authenticate(r)
-            return await self.list_tags(authenticate=False)
-        else:
-            msg = f"Unknown error listing tags from <{url}>: {r}"
-            raise DockerRegistryError(msg)
+        url = f"https://{registry}/v2/{repository}/tags/list"
```
athornton, line 133: Should we use

https://github.com/distribution/distribution/blob/main/reference/reference.go

and its nomenclature?  Maybe not.  But the terminology all gets very confusing and overloaded with "what's a registry, what's a repository, what's a reference, what's a path?"

src/jupyterlabcontroller/storage/docker.py
```diff
@@ -1,202 +1,313 @@
...
-        Returns the digest as a string, such as "sha256:abcdef"
+        Returns
+        -------
+        str
+            The digest, such as ``sha256:abcdef``.
         """
-        url = f"https://{self.host}/v2/{self.repository}/manifests/{tag}"
-        r = await self.http_client.head(url, headers=self.headers)
-        if r.status_code == 200:
+        url = f"https://{registry}/v2/{repository}/manifests/{tag}"
```
athornton, line 164: I eagerly await the day T&S needs to do an HTTP rather than an HTTPS repository because reasons.

athornton (approved):
I feel like the whole Docker registry/repository nomenclature is needlessly overloaded and difficult, but that's not really our fault.

The namespace stuff I still find confusing, largely because I feel like maybe it ought to somehow come out of nublado-application and nublado-users.application in phalanx `environments/`.  Maybe I'm worrying about something that really doesn't matter though.

src/jupyterlabcontroller/services/prepuller/docker.py
rra: I've been having a hard time wrapping my mind around this. I think we care about the digests for everything we're prepulling in order to detect and replace stale images as you mention, and we care about mapping an alias to another tag for recommended and any other alias we want to give a special position in the menu. But I don't think we care about knowing which tags are aliases other than that, do we? The only place they're used is in the dropdown menu, which is just a list of every tag, and I don't think we care what is an alias and what isn't. We just list all the tags anyway.

This is probably a good stand-up topic, since there may be some subtlety here that I'm missing. But it feels weird to do hundreds of GET requests against the Docker API pretty frequently when we by and large don't use any of that data, only the digests for the much smaller set of prepulled images.

src/jupyterlabcontroller/storage/docker.py
rra: I think we should use registry for the API server domain name and port and repository for the path inside the registry (without including the registry hostname itself), since that seems to be the terminology used in the Docker world. See, for instance:

https://stackoverflow.com/questions/34004076/difference-between-docker-registry-and-repository

I do think we should use reference uniformly for the full path including registry, repository, and tag and/or digest, as recommended by the above code. We've been using path (and earlier url) for that, but reference seems to be the official term, so we should switch. I'll take care of that in a subsequent diff.

rra:
Let's also talk about the namespace stuff in a stand-up. I think there may be ways to make it clearer in Phalanx. I think we should let people pick whatever prefix they want for their lab environments as a values override, but in our Phalanx we should default to the namespace in which the Helm chart is deployed, and that's something we can probably do in the Helm chart. (I'm also wondering if the Helm chart should itself deploy an Argo CD application to serve as a parent for all the resources instead of having a separate dummy Helm chart for that, although that's a little weird.)

PR #14, opened 2023-02-11
DM-37933: Refactor context handling

rra:
Dependency injection for mock storage clients wasn't plumbed in at quite the right layer, which created cascading complexity problems through the test suite and in main.py. Restructure using the following principles:

- The test suite injects mock storage objects by passing them as optional parameters to the constructor of the ProcessContext object.

- ContextDependency manages the global ProcessContext. Normally, it constructs one during initialize(), but to allow for the test suite to build a custom one with dependency injection and inject it, it supports an override method that prevents initialize() from making a new ProcessContext.

- Starting background processing is the responsibility of the ContextDependency (again during initialize()), via the start() method on the ProcessContext. (Currently, an ugly hack is required to do user map reconciliation, to be addressed later.)

- The per-request RequestContext should contain a Factory to make components that can be created on the fly, and copies of any global singletons that handlers need access to. These are taken from the ProcessContext, and the Factory also gets a reference to the ProcessContext.

- All the functions to create new service objects live in the Factory and are called via it. They don't need to be reflected in the RequestContext object since its Factory member is public and doesn't need to be hidden.

- The ProcessContext should be recreated per-test to ensure that it doesn't share state between tests. This also cleaned up the lifetime issues that required forcing the event loop to have session scope.

The Gafaelfawr client cannot easily be mocked via dependency injection since it's created by the Factory, so handle it with patching for now until it can be converted to httpx mocking.

Where possible, tests use a simple Factory instead of the mocked RequestContext. Code forcing an initial reconcile of prepuller state has been moved into the test fixtures and out of the tests.


athornton (approved):
The cleanup of the hideous global stuff looks really good.  I feel less certain about exposing the factory and de-propertizing the methods that give you the objects you actually use from a Request+Process Context but I certainly trust @rra 's aesthetics over my own here.

And in general I agree with pushing the mocks down to the layer of individual HTTP calls when we can.  But it also feels that if we're mocking at that level then probably the mock fixtures themselves belong in one of our shared-library-repos, whether that's The Monorepo or Safir.  Or is that a bad idea for some reason I'm not seeing?

rra:
In terms of exposing the factory, my logic is that you have to expose *something* that creates the components you need to use. In the previous code, the context was effectively the factory: it exposed a bunch of properties to construct things for you. This layer of abstraction doesn't feel to me like it's adding anything except a code maintenance burden, since every factory method has a corresponding context method that just forwards to the factory method. So may as well expose the factory and let people call it directly.

The question of a property versus a method is mostly stylistic, but I prefer methods for a factory because ideally the factory is constructing something that doesn't already exist. The line between a property and a method is a bit hazy, but to me a property is exposing something that already exists (possibly in a slightly altered form), and a method is doing something or creating something. Also, just in general when I read other code it seems typical for factory methods to be methods and not properties.

For the mocks, yes, we definitely should move the mocks into Safir. I have been doing that as soon as two different packages use the mock, so haven't done that for the Docker mock yet since for right now only the lab controller uses it. But for instance Safir provides Kubernetes and Google Cloud Storage mocks.

PR #15, opened 2023-02-11
DM-37933: Overhaul tag parsing

rra:
Move StandaloneRSPTag to a new RSPImageTag class in a separate domain model that provides an abstract data type. The rest of the current models/tag.py code will move in a separate change. Rework the parsing code to hopefully be easier to follow and a bit more succinct.

Comparing incomparable tag types now behaves like comparing any other incomparable types.  The cost is a less specific and verbose error message, but given that this indicates a programming error, that seems fine for the additional simplicity.


src/jupyterlabcontroller/models/tag.py
```diff
@@ -529,25 +145,11 @@ def _rebuild(self, tags: List[RSPTag]) -> None:
...
-        # will sort them with most recent first, because of the tag type
-        # definitions in SQR-059.
-        for tag_type in RSPTagType:
-            new_tags[tag_type].sort(reverse=True)
-        # And flatten it out into a homogeneous list.
-        flat_tags: List[RSPTag] = list()
-        for k in new_tags:
-            flat_tags.extend(new_tags[k])
-        self._all_tags = flat_tags
+        new_tags = []
```
athornton, line 148: Wow.  This is SO much clearer than what I was doing.

athornton (approved):
I had some initial confusion with the refactoring, but I think it's that I didn't retain the initial PR summary: the reason only RSPTagType has moved to RSPImageType/Tag is that there's going to be a followup that moves the other Tag classes?

athornton:
And I don't know if this belongs in a common package, or something that is explicitly a SQR-059 implementation.  It may be that nothing outside the Lab Controller ever really needs to care about tag format and meaning, although it will be nice to be able to wave it at our hapless release engineer the next time they manage to come up with an RC Candidate whose tag somehow doesn't match any of the existing and documented formats.

rra:
Yes, indeed, there are follow-up diffs to also move RSPTag that are coming. The goal at the end of this is to create abstract data types for representing RSP images and image tags that we can lift into Safir down the road and reuse in other projects. The starting point will have just the methods required for the lab controller, but the idea is that they won't otherwise depend on any lab controller code. They're just a generic representation of a tag and an image.

PR #17, opened 2023-02-13
DM-37933: Mock Gafaelfawr with respx

rra:
Replace the patched mock Gafaelfawr storage class with respx mocking so the actual GafaelfawrStorageClient can be used. This required some changes to the test data format to include the tokens that map to particular users.

The get_scopes method was never called, so remove it and thus the need to mock the token-info endpoint.


tests/settings.py
```diff
@@ -122,7 +125,7 @@ def userdatas(self) -> List[UserData]:
         userinfos = self.userinfos
         lab_statuses = [x for x in LabStatus]
         pod_states = [x for x in PodState]
-        for idx, v in enumerate(userinfos):
+        for idx, v in enumerate(userinfos.values()):
```
athornton, line 128: Was this wrong before, but accidentally passing?

tests/settings.py
```diff
@@ -94,8 +94,11 @@ def canonicalize(self) -> None:
         self._canonicalized = True
 
     @property
-    def userinfos(self) -> List[UserInfo]:
-        return [UserInfo.parse_obj(x) for x in self.test_objects["user_info"]]
+    def userinfos(self) -> dict[str, UserInfo]:
+        return {
+            t: UserInfo.parse_obj(d)
+            for t, d in self.test_objects["user_info"].items()
+        }
```
athornton, line 101: Hah.  The one time I actually don't just dive for a dict, it would have been the right choice.

tests/settings.py
rra: No, it was correct, I changed the list to a dict because I needed the names of the tokens to reuse the same data structure for the Gafaelfawr mock. So that required this change to get back the original list.

athornton (approved):
This looks good.  One bit of confusion on my part.

tests/settings.py
athornton: Oh, OK, I see now how it's coupled to the next bit.

PR #18, opened 2023-02-13
DM-37933: Recognize unknown tags with a cycle

rra:
On Telescope and Site deployments where it is vital to prevent spawning images from the wrong SAL cycle, alias tags like recommended also have a cycle appended. Recognize those tags sufficiently to capture the cycle so that they're not excluded by cycle filtering when constructing the image menu.


athornton (approved):
Good catch.

PR #19, opened 2023-02-13
DM-37933: Simplify Gafaelfawr interaction and Kubernetes mocking

rra:
For several routes, the lab controller was asking Gafaelfawr for user information about a token only to determine the authenticated username. Take advantage of the `X-Auth-Request-User` header added by Gafaelfawr on all authenticated calls to extract the username from the header and avoid the Gafaelfawr call, for a small reduction in complexity.

In the spawn handler, which does have to interact with Gafaelfawr, pass the resolved user object into the lab manager instead of having it resolve the user again.

Add a user-status end-to-end handler test, which uncovered various problems with the mocking of the Kubernetes layer (mostly missing methods). The Kubernetes storage layer mock was almost entirely dummy functions, so rather than have to reproduce the API of the actual storage layer, take advantage of `unittest.mock.Mock` and its spec argument to construct a mock with knowledge of the underlying API, and then selectively override only the methods that need to do something other than silently succeed.

Fix a couple of places where routes were registered on `"/"` instead of `""`, which adds an unwanted trailing slash to the URL and forces a FastAPI redirect. The test case ran into this because httpx doesn't follow redirects by default.


tests/handlers/user_status_test.py
```diff
@@ -0,0 +1,79 @@
...
+    config: Configuration,
+    mock_gafaelfawr: MockGafaelfawr,
+    obj_factory: TestObjectFactory,
+) -> None:
+    for token, user in obj_factory.userinfos.items():
+        break
+    lab = obj_factory.labspecs[0]
+    size_manager = SizeManager(config.lab.sizes)
+
+    # At the start, we shouldn't have any lab..
```
athornton, line 28: Nit: Either one or three periods here, but not two.

athornton (approved):
I like the header simplification very much.  I should probably learn my way around the mock library, shouldn't I?

rra:
Yeah, the mock library is incredibly useful.

PR #20, opened 2023-02-14
DM-37933: Add RSPImageTag constructor for alias tags

rra:
Add a new class method to create alias tags. Do a bit of parsing of the tag name to extract a cycle if present, ensuring that the tags won't be filtered out by a cycle restriction (and making the display name a bit nicer because we can).

This constructor will be used in an upcoming change.


athornton (approved):
This is one of those obviously-right why-didn't-I-think-of-that changes.

PR #21, opened 2023-02-14
DM-37933: Simplify Gafaelfawr interactions

rra:
Now that only lab creation has to contact Gafaelfawr, remove the token management from ContextDependency, remove get_user from RequestContext, and treat GafaelfawrClient like any other service that a handler may need that is created on demand by the handler using a factory.

Remove all the caching of Gafaelfawr replies. We only query Gafaelfawr on lab creation, which is already an expensive operation, so this optimization shouldn't be necessary and could risk out-of-date user information.

The lab tests were also relying on get_user in the request context, but with an additional helper funciton to get a user can now be rewritten to use a factory. Take advantage of that to simplify their code a bit and remove the now-unused user_context fixture.


athornton (approved):
I think this is OK.  It's requiring some adjustment in my head because apparently my own mental model is pretty inflexible that the user object is the thing being passed around, not something I just produce upon demand, but this is certainly simpler.

PR #22, opened 2023-02-15
DM-37933: Simplify create_app

rra:
Ensure all tests run with a mocked Kubernetes API. This is not currently used, since we currently also mock the Kubernetes storage object, but it ensures that initialize_kubernetes doesn't fail if there is no Kubernetes configuration. This allows removal of a workaround in create_app.

Move the startup and shutdown handlers into create_app to make it more explicit that they're private to create_app. Update the docstring with the true reason why we're deferring app creation: we want to give the test suite a chance to replace the configuration first.


athornton (approved):
Much cleaner.

PR #23, opened 2023-02-15
DM-37933: Include build information when sorting tags

rra:
semver ignores the build component of the version when comparing versions, but that's where we put the cycle and cycle build information. We therefore don't want to ignore it because we want tags with a newer cycle or cycle build to sort after ones with an older cycle or cycle build for otherwise identical versions, and tags with a cycle to sort after ones without a cycle. This ensures that when presenting tags in reverse sorted order, ones with a newer cycle or cycle build will be shown first.

### DIFF
--- a/src/jupyterlabcontroller/models/domain/rsptag.py
+++ b/src/jupyterlabcontroller/models/domain/rsptag.py
@@ -363,8 +363,23 @@ class RSPImageTag:
             return NotImplemented
         if self.image_type != other.image_type:
             return NotImplemented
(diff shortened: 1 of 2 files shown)


athornton (approved):
Thanks!  Yeah, I missed this case, although I think it is not problematic in practice because the display menu only displays the "correct" cycle tags and hides everything else.

Although I haven't tested the new spawner on a T&S site so I might be wrong about that.  Hmmm.  Anyway, this is an improvement.

PR #24, opened 2023-02-16
DM-37933: Add abstract data type collecting RSPImageTag

rra:
Analysis of available images and prepull state is very complex when expressed as operations on basic Python data types. In an attempt to divide that complexity into managable parts, introduce an abstract data type to collect RSPImageTags, return them in sorted order, filter them by cycle, and extract useful subsets of them.

The callers of this code will be added in subsequent changes. This is not ideal for review purposes, but the change that uses this data type is too large to be reviewable without extracting some preliminaries as separate changes. The supported operations here are only those needed by the subsequent change.


athornton (approved):
LGTM.  I particularly like the subset() helper.

PR #25, opened 2023-02-16
DM-37933: Convert all Docker mocking to respx

rra:
Set up respx mocks for the Docker API by default and remove the mocked Docker storage client. Change the API of the Docker storage client to take a path to the credentials file and manage it internally rather than taking the DockerCredentialsStore object, since that object isn't used outside of the Docker client.

This allows simplifications to the Factory and ProcessContext classes. Also remove the unused create method on the Factory, which I originally copied from Gafaelfawr, since it isn't needed here (standalone can always be used).


athornton (approved):
Looks like a good simplification.

PR #26, opened 2023-02-20
DM-37933: Add RSPImage abstract type and its collection

rra:
Continuing to lay the groundwork for a clearer expression of the complex image analysis algorithm, this introduces abstract data types for tracking images. An image is a tag annotated with its registry, repository, and digest, and is used to drive prepulling and to compare against when determining what images are cached.

As with RSPImageTagCollection, the operations on RSPImageCollection are chosen to satisfy the requirements of subsequent changes. This is where nearly all logic for alias handling, including resolving aliases to their underlying tags, is done.

### DIFF
--- a/src/jupyterlabcontroller/models/domain/rspimage.py
+++ b/src/jupyterlabcontroller/models/domain/rspimage.py
@@ -1,0 +1,331 @@
+"""Abstract data types for handling RSP images."""
+
+from __future__ import annotations
+
+from collections import defaultdict
+from collections.abc import Iterable, Iterator
+from dataclasses import asdict, dataclass, field
+from typing import Optional, Self
+
+from .rsptag import RSPImageTag, RSPImageType
+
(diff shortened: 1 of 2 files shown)


src/jupyterlabcontroller/models/domain/rspimage.py
```diff
@@ -0,0 +1,331 @@
...
+
+    def add(self, image: RSPImage) -> None:
+        """Add an image to the collection.
+
+        Parameters
+        ----------
+        image
+            The image to add.
+        """
+        # If we're adding a non-alias image and we have unresolved aliases for
```
athornton, line 158: Oh, that's nice.  It should be rare, so the expensive reindex should not happen with much frequency.

athornton (approved):
I like the use of subtract, and the "eh, let's just reindex from scratch" approach.

PR #27, opened 2023-02-20
DM-37933: Reformat test data

rra:
Whenever possible, compare objects in pytest by converting them to dictionaries and comparing the dictionaries with assert. pytest will then show a rich diff if there are differences, rather than showing both strings as unhelpfully long lines.

Convert the text dumps of objects to JSON, and compare the JSON for this improved error reporting. For Kubernetes API models, clean up the results of `to_dict` by removing all `None` settings so that the serialized output is shorter, more readable, and easier to find settings in when updating the test. Remove the test for generating lab volumes since the full output of that method is also included in the pod spec, and thus can be tested by the test of the pod spec method.


src/jupyterlabcontroller/models/domain/prepuller.py
```diff
@@ -5,6 +5,8 @@
 from dataclasses import dataclass, field
 from typing import Dict, List, Optional, TypeAlias
 
+from pydantic import BaseModel
```
athornton, line 8: Isn't BaseModel a strong hint that it's not a domain model?  I had thought in general Pydantic was to be used for things that will be part of the external API, and Dataclass for internal models.

tests/services/lab_test.py
```diff
@@ -86,17 +108,222 @@ async def test_env(
...
     std_result_dir: Path,
 ) -> None:
     _, user = obj_factory.get_user()
     lab_manager = factory.create_lab_manager()
+    volumes = lab_manager.build_volumes(user.username)
 
-    vols = lab_manager.build_volumes(user.username)
-    vol_str = "\n".join([f"{x}" for x in vols])
-    check_file(vol_str, std_result_dir / "volumes.txt")
+    # Unfortunately, the serialization of Kubernetes API objects is huge (it
```
athornton, line 120: I get it but it feels less consistent.

athornton (approved):
I am less philosophically on-board with this one than with most of the earlier ones, but I see why from a pragmatic standpoint this is better.

src/jupyterlabcontroller/models/domain/prepuller.py
rra: This is going to be temporary anyway since I think we can test the same things through the API, but I went ahead and put this back to a dataclass so that we can keep that invariant.

tests/services/lab_test.py
rra: This was a good prompt for me to think more about this, and I came up with a serialization format that I think will work, so now hopefully we're back to consistency.

athornton (approved):
I like that a lot better; strip_none() is an excellent idea.

PR #29, opened 2023-02-23
DM-37933: Ensure all labs are spawned with a digest

rra:
The decision to not retrieve digests for all available images means that labs spawned from the dropdown menu may not have digests. However, we always want to record the digest of the image when spawning the pod for debugging purposes and better user display. The lab service also had some left-over reference and tag parsing code to reconstruct information about the image from the reference passed in via the lab creation form.

Correct this by adding a new method to the image service to create an RSPImage from a Docker reference, and then use the RSPImage as the source of data to encode in the pod environment ConfigMap. In the process, change image to reference in the lab options model and refactor reference parsing so that it can be shared between the lab service and cached image analysis.

### DIFF
--- a/src/jupyterlabcontroller/exceptions.py
+++ b/src/jupyterlabcontroller/exceptions.py
@@ -1,3 +1,9 @@
+class InvalidDockerReferenceError(Exception):
+    """Docker reference is not valid."""
+
+    pass
+
+
 class DockerRegistryError(Exception):
     """Unknown error working with the Docker registry."""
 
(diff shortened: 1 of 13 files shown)


athornton (approved):
I like this.  I don't know why I didn't realize we could always resolve the digest just-before-pull, since if we don't already have the digest, we believe we don't have the image, and the HEAD is much cheaper than the pull.  (Or course we might already have the image, but I guess we know that as soon as we do the HEAD, and we do the pull anyway, it just might be free if we already have all the layers.)

PR #30, opened 2023-02-23
DM-37933: Switch to full Kubernetes API mocking

rra:
Stop using the mock Kubernetes storage object in favor of a full mock of the Kubernetes API for the methods that the lab controller calls. This is temporarily implemented as a subclass of the mock class provided by Safir, but those changes will be rolled back into Safir once they have been thoroughly exercised.

This uncovered a timing problem with one lab service API test, where deletion was handled by a background task and therefore the user map didn't update immediately. Handle that with a delay that should be more than long enough for the background deletion to clean things up.

Remove some code from the Kubernetes storage layer that is handled by the Watch API. Setting _preload_content isn't an optimization; it tells the API to return the raw HTTP response rather than the deserialized object, and is always set internally by the Watch API (and should never be set otherwise). There is similarly no need to provide resource_version unless we're resuming a previous watch, since Watch handles setting it automatically.

### DIFF
--- a/src/jupyterlabcontroller/dependencies/context.py
+++ b/src/jupyterlabcontroller/dependencies/context.py
@@ -71,7 +71,6 @@ class ContextDependency:
 
     def __init__(self) -> None:
         self._process_context: Optional[ProcessContext] = None
-        self._overridden = False
 
     async def __call__(
         self,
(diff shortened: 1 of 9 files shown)


athornton (approved):
The K8s mock is really nice.  I guess the idea is we just keep gluing on kubernetes_asyncio methods as we use them and need to mock them and the Safir mock gets richer and richer?

rra:
Yup, exactly. The whole API is too big to implement at once, but we can just keep adding functionality as we need it.

PR #31, opened 2023-02-23
DM-37933: Support bot user interface to lab creation

rra:
Implement the alternate interface to the /create route that allows services to select their image by ignoring the JupyterHub form and instead POSTing known values for image_tag or image_class, allowing them to bypass an API request to the lab controller to get the correct image reference to use.

Collapse the models for lab options into a single model and use validators to convert from the wire format resulting from the JupyterHub form post. This enlists Pydantic to do all the cleanup and validation for us and generate the correct 422 responses if needed, which simplifies the error handling and makes the models less complex and more consistent.

Update the username and group regexes from Gafaelfawr.


athornton (approved):
Ship it.

PR #33, opened 2023-03-08
DM-38279: Fix handling of image selection in lab UserOptions

rra:
Treat empty lists in UserOptions input equivalent to None. Add a test for UserOptions to ensure that DROPDOWN_SENTINEL_VALUE is handled correctly, lists are converted to regular elements, and booleans are converted correctly.

If both image_list and image_dropdown are set (but only those two), use image_list and ignore image_dropdown. Due to how our form is designed, image_dropdown will always be set, but if image_list is not the sentinel value, it should override.

Stop configuring UserOptions with use_enum_values. This was an attempt to put strings into the dict() output for easier testing, but the test suite has been refactored to not need this, and this put the string directly into the model attribute as well, breaking other parts of the code.

### DIFF
--- a/src/jupyterlabcontroller/models/v1/lab.py
+++ b/src/jupyterlabcontroller/models/v1/lab.py
@@ -178,6 +178,16 @@ class UserOptions(BaseModel):
         else:
             raise ValueError(f"Invalid boolean value {v}")
 
+    @validator("size", pre=True)
+    def _validate_size(cls, v: Any) -> Any:
+        """Lab sizes may be title-cased, so convert them to lowercase."""
+        if isinstance(v, LabSize):
+            return v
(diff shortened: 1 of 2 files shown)


athornton (approved):
This looks good, but can I request one validator addition?

When we ingest sizes from the incoming lab form, can we do something like options["sizes"] = [x.lower() for x in in options["sizes"]] ?  The reason for this is that we title-case the size when presenting the form to the user, and it would be nice to just hand the formdata straight back to the controller without doing any processing on the restspawner side.

Alternatively we could change the sizes enum values to be title-cased in the first place, but it feels easier to just flatten case of inbound sizes.

rra:
Sure, that's easy enough to do!

PR #40, opened 2023-03-17
DM-37833: Get notebook quota information from Gafaelfawr

rra:
Gafaelfawr now optionally includes quota information for the Notebook Aspect of the Science Platform if it is configured to track user quotas. Get the quota information for the namespace quota from there instead of from JupyterHub, which will never know what quota to use in the nublado design.

### DIFF
--- a/src/jupyterlabcontroller/models/v1/lab.py
+++ b/src/jupyterlabcontroller/models/v1/lab.py
@@ -212,9 +212,6 @@ class LabSpecification(BaseModel):
     env: dict[str, str] = Field(
         ..., title="Environment variables from JupyterHub"
     )
-    namespace_quota: Optional[UserResourceQuantum] = Field(
-        None, title="Quota for user"
-    )
 
 
 """GET /nublado/spawner/v1/labs/<username>"""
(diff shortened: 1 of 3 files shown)


athornton (approved):
I presume eventually we would have some way to assign quotas based on group membership or something similar?

Looks fine.

src/jupyterlabcontroller/models/v1/lab.py
```diff
@@ -237,6 +234,36 @@ class UserGroup(CamelCaseModel):
     )
 
 
+class NotebookQuota(BaseModel):
+    """Notebook Aspect quota information for a user."""
```
athornton, line 238: Might want to name this, or explain in the comment, to make it clear that this is the quota for the namespace, which is distinct from CPU/memory requests/limits (but the namespace quota must be at least as large as the CPU/memory limit)

src/jupyterlabcontroller/services/lab.py
```diff
@@ -463,9 +461,15 @@ async def create_quota(
             )
 
     def build_namespace_quota(
-        self, lab: LabSpecification
+        self, user: UserInfo
     ) -> Optional[UserResourceQuantum]:
-        return lab.namespace_quota
+        if user.quota and user.quota.notebook:
```
athornton, line 466: Will this let us separate namespace and pod limits in its current form?

src/jupyterlabcontroller/models/v1/lab.py
rra: This model is copied from Gafaelfawr, which thinks of this as the quota for the entire Notebook Aspect. I think this problem will go away once we are importing the Gafaelfawr models from a separate package. Added to my to-do to clean up how the user metadata is shown in the lab, though, since right now this is a bit awkward.

src/jupyterlabcontroller/services/lab.py
rra: No, Gafaelfawr doesn't support that. If that's something we'll want, we should talk about what that would look like. I'll add that to my queue for stand-up discussion.

PR #43, opened 2023-03-23
DM-38279: Wait for lab deletion before returning

rra:
If we wait for namespace deletion in the background and return immediately to JupyterHub, JupyterHub assumes that we have deleted the lab and may immediately attempt to create a new one, which then fails.

### DIFF
--- a/src/jupyterlabcontroller/handlers/labs.py
+++ b/src/jupyterlabcontroller/handlers/labs.py
@@ -105,7 +105,7 @@ async def post_new_lab(
         404: {"description": "Lab not found", "model": ErrorModel},
         403: {"description": "Forbidden", "model": ErrorModel},
     },
-    status_code=202,
+    status_code=204,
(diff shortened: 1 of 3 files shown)


athornton (approved):
This looks fine, although I picked the previous behavior on purpose to quickly get the user back to the landing page.  I thought the only thing that happened was that if you immediately clicked on the Hub again you'd get the "your server is stopping, please wait" message and then you'd have to wait out the 15 seconds or so tearing down your namespace takes.  If it's worse than that, sure, this makes plenty of sense.

rra:
What happened without this is that if mobu had to delete a lab at startup, the lab deletion would complete immediately and then it would try to create a new lab and that would immediately fail, so after each error it had a round of failing before it could successfully restart. Maybe there's something trickier we can do; I'll add this to my to-do list to look at in more detail.

PR #46, opened 2023-03-27
DM-38279: Fix quota handling

rra:
When I added support for Gafaelfawr quota settings, it broke pod spawning because I missed some pieces. Set resource requests and quotas on the pod based on the size of pod requested by the user, and set the same requests and limits for the init container (even though it probably won't need nearly as much) since that preserves the same total pod requirements and is easy.

Refactor the size manager to use methods rather than properties that do a lot of work, and to return the resources for a specific lab size, which is a bit more convenient elsewhere in the code.

### DIFF
--- a/src/jupyterlabcontroller/services/lab.py
+++ b/src/jupyterlabcontroller/services/lab.py
@@ -22,6 +22,7 @@ from kubernetes_asyncio.client.models import (
     V1PodSecurityContext,
     V1PodSpec,
     V1ResourceFieldSelector,
+    V1ResourceRequirements,
     V1SecretVolumeSource,
     V1SecurityContext,
     V1Volume,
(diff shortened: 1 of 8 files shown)


src/jupyterlabcontroller/services/lab.py
```diff
@@ -686,27 +687,29 @@ def build_volumes(self, username: str) -> List[LabVolumeContainer]:
...
             if ic.privileged:
                 ic_sec_ctx = V1SecurityContext(
                     run_as_non_root=False,
                     run_as_user=0,
                     allow_privilege_escalation=True,
                 )
+            else:
+                ic_sec_ctx = V1SecurityContext(
+                    run_as_non_root=True,
+                    run_as_user=1000,
```
athornton, line 708: Is there ever going to be a case when we want to run an unprivileged user container but we want to run it as the right user?  I couldn't think of one, because I think whatever it is we want to do, we probably want to do in the lab, but I could be convinced that we should plug the real uid in here.

athornton (approved):
This all looks good.

src/jupyterlabcontroller/services/lab.py
rra: Yeah, I think we probably should default to running init containers as the user. This was pre-existing, so I'll add that to my to-do list for later attention.

PR #47, opened 2023-03-28
DM-38279: Change order of progress messages

rra:
Send the new progress percentage before the message, since the REST spawner class needs the new progress percentage when it emits the message.


athornton (approved):
LGTM

PR #48, opened 2023-03-28
DM-38279: Return 201 from lab create route

rra:
The URL to which we redirected the lab creation response is protected by different authentication rules and would require a different token. This probably indicates a deeper problem that we should rethink, but since the REST spawner doesn't want to follow the redirect anyway, return 201 with an empty body and a Location header instead.


athornton (approved):
LGTM

PR #49, opened 2023-03-29
DM-38279: Don't overwrite pod labels

rra:
The standard Argo CD labels for the pod weren't present because they were overwritten by the extra label we wanted to add. Update the dict instead of overwriting it.


athornton (approved):
Obviously correct.

PR #51, opened 2023-03-29
DM-38279: Don't return an event stream for nonexistent users

rra:
If someone asks us for events for a user that we've never heard of, we'll just block forever unless someone happens to create a lab for that user. Instead, return 404 if we've never heard of the user. Labs should always be created before asking for their events, and if the lab doesn't exist, the events shouldn't either.


athornton (approved):
Obviously correct.

PR #52, opened 2023-03-29
DM-38279: Test suite improvements

rra:
Add end-to-end tests of the lab routes, and add a test for all of the Kubernetes objects created by lab spawn. This replaces the previous tests calling directly into the lab service, which only tested the pod, whereas this covers all of the objects.


src/jupyterlabcontroller/storage/k8s.py
```diff
@@ -359,6 +359,8 @@ async def create_secret(
         immutable: bool = True,
     ) -> None:
         secret = V1Secret(
+            api_version="v1",
```
athornton, line 362: kind and version are implicitly set, so I don't get why you're making them explicit, but OK I guess....

athornton (approved):
Simplification is good.

src/jupyterlabcontroller/storage/k8s.py
rra: Yeah, this will get fixed later. I was setting them solely for the test suite so that I have them available when testing the objects, but this work belongs in the mock layer rather than in the code. I'll move it in an upcoming PR. Sorry about that!

PR #53, opened 2023-03-29
DM-38279: Handler refactoring and testing

rra:
Do some wide-ranging small cleanups of the route handlers to make this application look closer to our others. The primary changes are:

- Include the full path to the route when registering it, apart from the prefix added to all routes. This is what we've been doing elsewhere except for Gafaelfawr, and I am going to change Gafaelfawr to do the same. It removes the need for comments in the handler files explaining what the full route path is.
- Use `Request.url_for` to get links to other resources rather than having to be configured with our external URL.
- Remove docstrings for handlers in favor of `title` and `description`, since otherwise FastAPI sometimes tries to mingle the docstring and the description and weird things happen.
- Remove response models that were no longer used.
- Rename `indices.py` to `index.py`; the former is technically more correct but the latter is what we've used elsewhere.

The Safir configuration now supports the standard environment variables and uses the same naming as times-square (so `pathPrefix` instead of `rootEndpoint`). Remove the logger name configuration, which we've been dropping from all of our applications.

Remove some dead code, including some duplicate routes that weren't used.

Tests are now by preference done end-to-end through the handler instead of by calling into the service layer and bypassing the handler layer, which will give us better coverage on the handlers.


athornton (approved):
This seems good, although I liked having the pointer to the particular part of sqr-066 that defined each response, because it makes it very easy to flip back and forth between "this is what the spec says I should return" and "this is what I am doing".

rra:
My thought process there was that at some point I think the lab controller documentation itself should take over as the canonical description of the API, rather than hand-writing an API description in a tech note as well. It's useful to have the initial API in the tech note, but I don't think we want to hand-maintain the tech note in the long run when Swagger can generate API documentation for us. That's what I did with Gafaelfawr; I wrote the initial API in the tech note, but now the canonical API description is in Gafaelfawr itself.

PR #54, opened 2023-03-29
DM-38279: Use a start script to start the application

rra:
Follow the pattern from Gafaelfawr, even though our start command is fairly simple. Add the --factory flag, which uvicorn warns is missing when not provided.

### DIFF
--- a/scripts/start.sh
+++ b/scripts/start.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+#
+# Start the lab controller application inside the Docker image.
+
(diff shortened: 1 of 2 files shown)


scripts/start.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
athornton, line 1: Does this need to be bash?  set -u is POSIX, isn't it?

rra: Good catch.

PR #55, opened 2023-03-29
DM-38279: Standardize type annotations and Pydantic models

rra:
Some general cleanup to typing and model names:

- Adopt the Python 3.11 type annotation conventions, stop using the deprecated typing aliases, and use `Self`.
- Now that the `Self` type avoids the conflicts between Pydantic and a class named `Config`, use `Config` for the configuration matching our other packages. Rename the other configuration models accordingly.
- Only use `CamelCaseModel` for configuration; the network API should always use snake_case.
- Simplify the enums; the special enum class didn't seem to save lines of code compared to initializing with the value and possibly inheriting from `str`.


src/jupyterlabcontroller/models/v1/event.py
```diff
@@ -1,24 +1,24 @@
...
 """GET /nublado/spawner/v1/labs/username/events"""
 
 
-class EventTypes(NubladoEnum):
-    COMPLETE = auto()
-    ERROR = auto()
-    FAILED = auto()
-    INFO = auto()
-    PROGRESS = auto()
+class EventTypes(Enum):
```
athornton, line 11: If we're using Python 3.11 we could use StrEnum instead, and then we could use auto() and still get the right behavior.

PR #56, opened 2023-03-29
DM-38279: Wait for first image service run during startup

rra:
We don't want to start answering user requests until we've populated our list of available images, since otherwise the spawner form may be bogus. Implement this by not returning from start until the first update pass is complete.

Don't start background services by default when using a test factory and update the tests that use it to start the background services themselves.


athornton (approved):
The factory as context manager took some getting my head around, but this seems like a good idea.

PR #57, opened 2023-03-30
DM-38279: Refactor event management

rra:
The previous way of handling events used a queue, which meant that only one reader was supported and if anyone else requested the pod events for a user, one or both of the listeners would miss events and possibly hang forever.

Rework the event manager to return separate iterators for every caller that use an asyncio Event to tell them when there are new events to read. Improve the event models a bit in the process. Add new tests and fix various issues in the mocks that those tests uncovered. Add a bit more logging that was helpful in figuring out why the test suite wasn't working.


src/jupyterlabcontroller/services/events.py
```diff
@@ -1,86 +1,136 @@
...
+        for trigger in triggers:
+            trigger.set()
+
+    def _remove_trigger(self, username: str, trigger: asyncio.Event) -> None:
+        """Called when a generator is complete.
+
+        Does some housekeeping by removing the `asyncio.Event` for that
+        generator. Strictly speaking, this probably isn't necessary; we will
+        release all of the events when the user is reset, and there shouldn't
+        be that may requests for the user's events before that happens for the
```
athornton, line 122: "that may" -> "enough"

src/jupyterlabcontroller/services/lab.py
```diff
@@ -94,31 +94,31 @@ async def info_event(self, username: str, message: str, pct: int) -> None:
...
-        ev_queue = self.event_manager.get(username)
+        event = Event(data=str(pct), type=EventType.PROGRESS)
+        self.event_manager.publish_event(username, event)
         umsg = f"{message} for {username}"
-        await ev_queue.asend(Event(data=str(pct), event=EventTypes.PROGRESS))
-        await ev_queue.asend(Event(data=umsg, event=EventTypes.INFO))
-        self.logger.info(f"Event: {umsg}: {pct}% ")
+        event = Event(data=umsg, type=EventType.INFO)
+        self.event_manager.publish_event(username, event)
+        self.logger.info(f"Event: {umsg}: {pct}%")
```
athornton, line 102: Helpful during debugging (which is why I had it there) but I bet you're gonna want this at loglevel.DEBUG before long

tests/handlers/labs_test.py
```diff
@@ -146,11 +180,121 @@ async def test_lab_start_stop(
...
+            {
+                "data": f"Autoscaling cluster for reasons for {user.username}",
+                "event": "info",
+            },
+            {"data": "60", "event": "progress"},
+            {
+                "data": f"Mounting all the things for {user.username}",
+                "event": "info",
+            },
+            {"data": "69", "event": "progress"},
```
athornton, line 278: Nice

athornton (approved):
LGTM; one comment that needs rewording, which feels churlish, since this is a huge improvement over the old state.

rra:
Oh, no, catching spelling errors and comment mistakes is great. Always happy to have the feedback!

PR #58, opened 2023-03-31
DM-38279: Only stop background services if started

rra:
The Factory object was unconditionally stopping background services when closed, but we don't always start them now. This produced spurious warning messages when looking at the logs from tests. Only stop the background services if we started them.


athornton (approved):
LGTM

PR #59, opened 2023-03-31
DM-38279: Also check the created namespace

rra:
Modify the Kubernetes mock to add the created namespace to its objects so that we can check it has the right annotations.


athornton (approved):
Still not a fan of explicit version/kind specification, but sure.

PR #60, opened 2023-03-31
DM-38279: Allow for empty dicts to docker or gar image config

rra:
Helm chart defaults make empty dicts nicer than null values, since it makes it clear the value should be a dictionary. Transform those into None in the validator.

Remove the unnecessary name annotations on the remaining configuration fields.


athornton (approved):
LGTM

PR #61, opened 2023-04-03
[neophile] Update dependencies

sqrbot:
- Update frozen Python dependencies
- Update psf/black pre-commit hook from 23.1.0 to 23.3.0


rra:
Fixed in https://github.com/lsst-sqre/jupyterlab-controller/pull/91

PR #62, opened 2023-04-03
DM-38466: Move more image operations into RSPImageCollection

rra:
With Google Artifact Registry, we can use an `RSPImageCollection` to hold all known images, rather than using the more limited `RSPImageTagCollection` because (with Docker) we don't have digests. `RSPImageCollection` therefore needs a `subset` operation to generate the list of prepulled images.

Alias tracking of images was incomplete: it didn't handle the case of multiple aliases pointing to the same image or non-alias tags with the same digest. The flags to `all_images` also hid images based on whether we had ever seen an alias even if that alias wasn't in the same collection, which wasn't desirable. Fix both of those problems by adding more sophisticated alias tracking to `RSPImageCollection`.

Similarly, the code for updating which nodes an image was present on was not sophisticated enough to understand alias relationships. Move that into `RSPImageCollection` so that it can share the same alias chasing logic.

This PR only updates the abstract data type but doesn't yet use the new functionality. That will be coming in the next PR that adds Google Artifact Registry support and refactors the image service.

### DIFF
--- a/src/jupyterlabcontroller/models/domain/rspimage.py
+++ b/src/jupyterlabcontroller/models/domain/rspimage.py
@@ -185,35 +185,10 @@ class RSPImageCollection:
                 return
 
         # This is a compact version of the same logic as _replace_contents.
-        if image.is_possible_alias and image.digest in self._by_digest:
+        if image.digest in self._by_digest:
             other = self._by_digest[image.digest]
-            for alias in other.aliases:
-                if alias == image.tag or alias not in self._by_tag_name:
-                    continue
-                self._by_tag_name[alias].aliases.add(image.tag)
-            if other.image_type in _ALIAS_TYPES:
-                image.aliases.add(other.tag)
-                other.aliases.add(image.tag)
-                self._unresolved_aliases[image.digest].append(image)
-            else:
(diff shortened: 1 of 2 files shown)


src/jupyterlabcontroller/models/domain/rspimage.py
```diff
@@ -311,21 +435,76 @@ def _replace_contents(self, images: Iterable[RSPImage]) -> None:
...
         for image in images:
-            self._by_tag_name[image.tag] = image
             self._by_type[image.image_type].append(image)
 
         # Sort all of the images by reverse order so the newest are first.
         # This allows all_images to return sorted order efficiently.
         for image_list in self._by_type.values():
             image_list.sort(reverse=True)
+
+    def _resolve_duplicate(self, new: RSPImage, old: RSPImage) -> None:
```
athornton, line 453: Thank goodness you not only unravelled all of this but documented the application attempts.

athornton (approved):
I very much like having made this explicit and laid out the steps of deduplication.

Life would have been a lot simpler if Docker just gave us the top-level digest in the first place for any image inquiry.

PR #63, opened 2023-04-03
DM-38466: Add Google Artifact Registry support

rra:
Add support for using Google Artifact Registry to discover remote images, alongside support for a Docker Registry.

This required introduction of a new layer of abstraction, an image source, with a shared API but different under-the-hood implementations. This takes over from the image service the process of analyzing remote images and updating them with information about which nodes they're cached on, and also handles some of the information requests to the image service. The image service is now responsible for managing the background update thread and calculating which images should be prepulled on which nodes, as well as some of the work for constructing the prepuller status.

Add more comprehensive tests for the prepuller and a new set of utility functions to make it easier to use test data for non-default test configurations.

Rework how image sources are specified to be friendlier to Pydantic verification. There is now a single source key that contains the source configuration, and Google Artifact Registry versus Docker Registry are selected via a `type` key. Plumb those new data types down into the storage layer, which makes the API for some of the storage calls simpler.

Close the Kubernetes API client properly on test teardown to avoid annoying warnings from pytest.


src/jupyterlabcontroller/models/v1/prepuller_config.py
```diff
@@ -20,60 +35,108 @@ class DockerDefinition(CamelCaseModel):
     )
 
 
-class GARDefinition(CamelCaseModel):
+class GARSourceConfig(CamelCaseModel):
+    """Google Artifact Registry from which to get images.
+
+    The Google Artifact Repository naming convention is unfortunate. It uses
```
athornton, line 41: Argh.  (Not you, Google)

src/jupyterlabcontroller/services/image.py
```diff
@@ -430,6 +345,39 @@ async def prepuller_wait(self) -> None:
...
+        node_cache
+            Mapping of node names to the list of cached images on that node.
+
+        Returns
+        -------
+        dict of str to RSPImageCollection
+            Image collections of images found on each node.
+        """
+        image_collections = {}
+        for node, node_images in node_cache.items():
```
athornton, line 369: If you're into abusing comprehensions, you could do this as a nested pair of list comprehensions.  Please don't though.

src/jupyterlabcontroller/services/source/base.py
```diff
@@ -0,0 +1,135 @@
...
+from ...models.domain.form import MenuImage
+from ...models.domain.kubernetes import KubernetesNodeImage
+from ...models.domain.rspimage import RSPImage, RSPImageCollection
+from ...models.v1.prepuller import PrepulledImage
+from ...models.v1.prepuller_config import PrepullerConfig
+
+__all__ = ["ImageSource"]
+
+
+class ImageSource(metaclass=ABCMeta):
```
athornton, line 20: Neat.  I like the idea of an abstract superclass that these things inherit from.  That will also make it easy if we determine that ghcr has a better-than-Docker API or something.

src/jupyterlabcontroller/services/source/gar.py
```diff
@@ -0,0 +1,201 @@
...
+
+__all__ = ["GARImageSource"]
+
+
+class GARImageSource(ImageSource):
+    """Image source using a Google Artifact Registry.
+
+    Google Artifact Registry has a much nicer API that returns all available
+    images with their aliases and hashes in one request. This means we don't
+    have to track as complex of data.
```
athornton, line 27: This sentence missing words.  Or at least "as complex of data" doesn't read very well.

tests/handlers/labs_test.py
```diff
@@ -52,46 +51,6 @@ async def get_lab_events(
     return events
 
 
-def strip_none(model: dict[str, Any]) -> dict[str, Any]:
```
athornton, line 55: Awww, so long, "strip_none".  I hardly knew you, but I liked you.

tests/support/gar.py
```diff
@@ -0,0 +1,87 @@
...
+        Parameters
+        ----------
+        parent
+            Parent search key that should uncover this image.
+        image
+            Image to add.
+        """
+        self._images[parent].append(image)
+
+    async def list_docker_images(
```
athornton, line 42: Yuck.  Figuring out the typing for this would have been very hard for me.

tests/support/kubernetes.py
```diff
@@ -39,7 +39,51 @@
...
 
-__all__ = ["MockLabKubernetesApi", "patch_kubernetes"]
+__all__ = [
+    "MockLabKubernetesApi",
+    "patch_kubernetes",
+    "strip_none",
+]
+
+
+def strip_none(model: dict[str, Any]) -> dict[str, Any]:
```
athornton, line 49: Oh, it just moved over here.  Cool.  Welcome back, "strip_none"!

athornton (approved):
Generally looks good...but maybe we really *do* want to think about splitting the image handling into a separate library and adding image deletion methods, which would give us an easy way to put the image purger back (it went away with Nublado v2).

src/jupyterlabcontroller/services/source/__init__.py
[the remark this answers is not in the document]
rra: I'm certainly not opposed, but let's get Nublado v3 out the door first and then we can look at that when we look at image pruning again. I'll add it to my to-do.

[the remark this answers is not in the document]
rra: (Not to imply that I'm going to be the one to do it, just saying I'll try to make sure we don't lose track of this.)

tests/configs/standard/input/config.yaml
```diff
@@ -30,100 +12,41 @@ lab:
...
     JUPYTERHUB_DEFAULT_URL: /lab
-    JUPYTERHUB_OAUTH_CLIENT_ALLOWED_SCOPES: '[]' # needed?
-    NO_ACTIVITY_TIMEOUT: "432000"  # Also from group?
+    JUPYTERHUB_OAUTH_CLIENT_ALLOWED_SCOPES: '[]'
+    NO_ACTIVITY_TIMEOUT: "432000"
     NO_SUDO: "TRUE"
     S3_ENDPOINT_URL: "https://storage.googleapis.com"
     SODA_ROUTE: /api/image/soda
     TAP_ROUTE: /api/tap
-    # WORKFLOW_ROUTE: /wf  # No longer used
```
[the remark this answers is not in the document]
rra: Yeah, good point, I think it belongs in separate documentation but we don't have that yet so I'll bring this back for now until that documentation exists.

PR #64, opened 2023-04-04
DM-38533: Merge the Kubernetes mock with Safir

rra:
In preparation for lifting the new version of the mock up into Safir, merge it with the Safir mock rather than subclassing. Fix a few places where the lab controller mock was using unnecessary test functions when there was a standard API function that did the same thing, and fix a few places where Safir was special-casing certain types of objects unnecessarily. Add docstrings to everything just for the hell of it.

Canonicalize kind, api_version, and namespace on object storage. These are checked if already set but if not set they're set based on the API call, matching the behavior of the Kubernetes API. Remove all the places where they were set in the lab controller storage layer, which allows simplifying the code that constructs object metadata.

### DIFF
--- a/src/jupyterlabcontroller/services/lab.py
+++ b/src/jupyterlabcontroller/services/lab.py
@@ -475,7 +475,7 @@ class LabManager:
     async def create_user_pod(
         self, user: UserInfo, resources: UserResources, image: RSPImage
     ) -> None:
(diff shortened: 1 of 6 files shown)


athornton (approved):
With the understanding I didn't work too hard at it (since it needs further refactoring), seems fine.

PR #65, opened 2023-04-04
DM-38554: Adjust environment variable setup

rra:
Also set CPU_GUARANTEE, which is part of the normal environment variables set by JupyterHub when it knows the resource limits. Comment JUPYTER_IMAGE, which we unfortunately have to keep setting for now. Drop EXTERNAL_UID, EXTERNAL_GID, and EXTERNAL_GROUPS, which were only for old labs that did the templating of /etc/passwd internally.

Trim from the standard configuration some environment variable settings that are no longer part of the standard configuration in Phalanx, and trim environment variables normally provided by JupyterHub that we pass on verbatim and thus wouldn't appear in the configuration. This makes the test configuration a bit closer to a Phalanx configuration.


athornton (approved):
LGTM.  I can change our lab extensions to use the non-deprecated variables pretty trivially.

PR #66, opened 2023-04-04
DM-38554: Move ACCESS_TOKEN out of the ConfigMap

rra:
Inject ACCESS_TOKEN into the environment directly from the secret rather than putting a copy of it into the environment ConfigMap.


athornton (approved):
LGTM

PR #68, opened 2023-04-04
DM-38554: Rework how pull secrets are implemented

rra:
Rather than listing the pull secret like any other secret and then separating it out if it's named pull-secret, use a separate configuration option specifying the pull secret to copy and explicitly copy it in the lab service independent of creation of other lab secrets.


athornton (approved):
LGTM

PR #70, opened 2023-04-04
DM-38554: Rename K8S_NODE_NAME to KUBERNETES_NODE_NAME

rra:
Avoid cryptic abbreviations in environment variables that are exposed to users.


athornton (approved):
LGTM

PR #71, opened 2023-04-04
DM-38554: Use structured exceptions for expected errors

rra:
Previously, the handlers were checking for expected errors and turning them manually into `HTTPException` exceptions. The detail part of those exceptions was generally just a string, rather than the error model the API specification claimed we were using, and each new exception had to be manually caught.

Import an idea from Gafaelfawr where there is a top-level exception class with a global exception handler that understands how to format itself into the error model using class variables that can be set by subclassed exceptions to indicate things like the error code and the HTTP status. Convert all the expected exceptions that should result in normal error codes (not a 500) to be based on that exception class. Annotate the exceptions on the way through with location information about the input that caused the error, where appropriate.

### DIFF
--- a/src/jupyterlabcontroller/exceptions.py
+++ b/src/jupyterlabcontroller/exceptions.py
@@ -121,6 +121,13 @@ class PermissionDeniedError(ValidationError):
     status_code = status.HTTP_403_FORBIDDEN
 
 
+class UnknownUserError(ValidationError):
+    """No resource has been created for this user."""
+
+    error = "unknown_user"
+    status_code = status.HTTP_404_NOT_FOUND
+
+
 class UnknownDockerImageError(Exception):
     """Cannot find a Docker image matching the requested parameters."""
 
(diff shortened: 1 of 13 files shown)


src/jupyterlabcontroller/exceptions.py
```diff
@@ -1,81 +1,176 @@
...
+    "NSCreationError",
+    "NSDeletionError",
+    "UnknownDockerImageError",
+    "UnknownUserError",
+    "ValidationError",
+    "WaitingForObjectError",
+]
+
+
+class ValidationError(SlackIgnoredException):
```
athornton, line 28: That's cool.

src/jupyterlabcontroller/main.py
```diff
@@ -65,4 +67,12 @@ async def shutdown_event() -> None:
         await context_dependency.aclose()
         await http_client_dependency.aclose()
 
+    @app.exception_handler(ValidationError)
```
athornton, line 70: That's also cool.

athornton (approved):
This is a really nice error-handling framework.

PR #72, opened 2023-04-05
DM-38554: Add a test for init containers

rra:
Use the same init container specification as IDF dev. Also sort the standard test configuration by key.


athornton (approved):
LGTM

PR #73, opened 2023-04-05
DM-38554: Switch to JSON events

rra:
My original idea to use separate events for progress and messages to avoid adding another encoding layer turned out to be a poor one. This required worrying about the ordering of messages and more complex logic on both sides, and the httpx-sse library supports decoding a JSON body of a server-sent event. Switch to JSON-encoded messages that include both a message and an optional progress value.


athornton (approved):
Yeah, this works a lot more cleanly.

PR #75, opened 2023-04-11
DM-38554: Add a test for cycle handling

rra:
Test that cycles are enforced and images from other cycles are omitted from prepull status and from the menu, including the dropdown.

This uncovered various issues with cycle tracking in aliases and experimental images, and a few annoyances in how display names were generated. Fix those up and add additional tests.


src/jupyterlabcontroller/models/domain/rspimage.py
```diff
@@ -132,10 +132,29 @@ def resolve_alias(self, target: RSPImage) -> None:
...
+        # parenthetical, so do that and otherwise keep the cycle-aware display
+        # name parsing done in the RSPImageTag alias method.
+        if " (" in self.display_name:
+            cutoff = self.display_name.index(" (")
+            base_display_name = self.display_name[:cutoff]
+        else:
+            base_display_name = self.tag.replace("_", " ").title()
+
+        # If the target has a SAL cycle, it already has parentheses in its
+        # description. Nested parentheses are ugly, so convert that to another
```
athornton, line 150: Some of us like Lisp more than others, but, sure, if it bugs you, go for it.

athornton (approved):
This looks nice.

PR #76, opened 2023-04-11
DM-38554: Update pod state in user status during spawn

rra:
This is a temporary hack so that the pod state is not completely wrong. The status request should check to see if the pod is there rather than just trusting it is based on internal state, but that's not a release blocker and thus a problem for another day.


athornton (approved):
Yeah, not sure how I missed doing this on the initial pass.  Clearly correct.

PR #77, opened 2023-04-11
DM-38554: Improvements to lab spawn messages

rra:
Sending separate progress events for the creation of every type of object spams the event list and tends to push the interesting events towards or off the bottom of the screen. Since all those events are identical every time, only emit events for creating the namespace and the supporting resources.

Tweak the percentages based on my subjective impression of what steps take the most time, leaving more room for the pod spawning itself. Use 75% as the threshold at which we hand control back to JupyterHub and wait for the lab to start.

Remove the username from every progress message. Since the progress messages are only returned by calling a user-specific endpoint, this seems pointless. Instead, only mention the user in the first and last messages.

### DIFF
--- a/src/jupyterlabcontroller/services/lab.py
+++ b/src/jupyterlabcontroller/services/lab.py
@@ -96,34 +96,32 @@ class LabManager:
         r = self.user_map.get(username)
         return r is not None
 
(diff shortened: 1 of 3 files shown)


athornton (approved):
LGTM

PR #78, opened 2023-04-11
DM-38554: Reduce severity of state recreation log message

rra:
When reconstructing internal state from Kubernetes after a restart, it's normal to encounter user labs that are not present in the internal state. Log this at info level rather than warning.


athornton (approved):
LGTM

PR #79, opened 2023-04-12
DM-38554: Do not include Gafaelfawr headers in schema

rra:
Various routes of the lab controller pull information from the X-Auth-Request-User and X-Auth-Request-Token headers, set by Gafaelfawr. By default, FastAPI includes that in the OpenAPI schema, but in this case this is an internal implementation detail of our authentication system and the client should never send these headers. Mark them as not included in the schema.


athornton (approved):
LGTM

PR #80, opened 2023-04-12
DM-38554: Add ownership information to prepull pods

rra:
Use the injected pod metadata for the lab controller to mark the prepull pods as owned by the lab controller. Set block owner deletion to try to force cleanup of the prepull pods before restarting the lab controller, to reduce conflicts after restart.

### DIFF
--- a/src/jupyterlabcontroller/services/prepuller.py
+++ b/src/jupyterlabcontroller/services/prepuller.py
@@ -2,10 +2,11 @@
 
 import asyncio
 import re
+from pathlib import Path
 from typing import Optional
 
 from aiojobs import Scheduler
-from kubernetes_asyncio.client import V1Container, V1PodSpec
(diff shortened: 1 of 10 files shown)


src/jupyterlabcontroller/services/prepuller.py
```diff
@@ -187,3 +199,29 @@ def _prepull_pod_name(self, image: RSPImage, node: str) -> str:
 
         # Kubernetes object names may be at most 253 characters long.
         return name[:253]
+
+    def _prepull_pod_owner(self) -> V1OwnerReference | None:
```
athornton, line 203: That's very clever.  I didn't realize that the owner reference would do that.

athornton (approved):
I didn't realize owner references would do that.  Nice.

PR #81, opened 2023-04-12
DM-38554: Fix cycle handling with Google Artifact Registry

rra:
Cycle filtering wasn't implemented on RSPImageCollection or plumbed through to the Google Artifact Registry storage layer. Do that, and add a test suite for cycle handling with Google Artifact Registry as well.

This uncovered a few more capitalization problems when an unknown tag with a cycle is later resolved into an alias. Fix those by always title-casing the display name when resolving the alias.


athornton (approved):
LGTM, although practically I don't think T&S is ever likely to use GAR, and I don't think anyone but T&S is ever likely to use cycles.

PR #83, opened 2023-04-12
DM-38554: Add Slack-reportable structured web exceptions

rra:
Add a class that understands httpx exceptions and derive Gafaelfawr and Docker exceptions from that class. Separate out the Gafaelfawr exception from parsing the result in a separate exception that understands Pydantic errors. These new exception classes support Slack reporting, which is not yet used but will be in an upcoming change.


athornton (approved):
Awesome

PR #84, opened 2023-04-12
DM-38554: Add structured exception for Kubernetes errors

rra:
Flesh out KubernetesError with the ability to report itself to Slack and the ability to create itself from an ApiException and capture details of that exception. Ensure all Kubernetes ApiException errors are converted to it. Be more restrictive in what Kubernetes errors we catch in the storage layer, since other unexpected errors should start as unexpected until we understand them.

Simplify the error handling for user state reconciliation to only skip past 404 errors and to abort the first time there's a fatal Kubernetes API exception. We will want to add retry logic at a higher level, but continuing without pause when Kubernetes is raising API exceptions is likely to be pointless.

### DIFF
--- a/src/jupyterlabcontroller/exceptions.py
+++ b/src/jupyterlabcontroller/exceptions.py
@@ -7,6 +7,7 @@ from typing import ClassVar, Optional, Self
 
 from fastapi import status
 from httpx import HTTPError, HTTPStatusError, RequestError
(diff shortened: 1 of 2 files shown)


athornton (approved):
LGTM.  The new Exceptions are definitely an improvement, and moving the recreation of a namespace into a helper function makes sense.

PR #85, opened 2023-04-13
DM-38554: Rework exception handling layer for lab spawning

rra:
Move all Kubernetes object creation into a background thread, rather than starting only with the lab creation, so that all error messages are reported via the events API and none have to be reported as a direct response from the route handler.

Standardize the exception handling during lab creation and deletion to happen at the top level and uniformly report failure events. Fix the event message on success events for deleting pods.

Add a stringification of KubernetesError and use the string value of the exception, without additional prefixes, as the error message to the user.

Fix the logic when waiting for pod creation if the lab hasn't been created yet.

Remove the now-obsolete services/lab_test test, since everything it tests is now tested through the handler.


src/jupyterlabcontroller/exceptions.py
```diff
@@ -379,7 +379,7 @@ def from_exception(
             namespace=namespace,
             name=name,
             status=exc.status,
-            body=exc.body,
+            body=exc.body if exc.body else exc.reason,
```
athornton, line 382: This syntax always seems Perlish to me.

athornton (approved):
LGTM.  I really like the new exception framework, and the simplification of task management helps.

PR #86, opened 2023-04-13
DM-38554: Set EXTERNAL_{UID,GID} for init containers

rra:
Our init container implementation depended on these variables, even though the main lab no longer does. Set them directly for init containers.


athornton (approved):
Yup.  Can you think of a cleaner way to do this for the init container?  No matter what we're gonna have to get those values in there, and I feel like a configmap would be more work.

rra:
Nah, I think this is the right approach. It's pretty simple and straightforward, and we can just document that those variables are set for init containers.

PR #88, opened 2023-04-14
DM-38554: Catch and report other prepuller errors

rra:
We were already handling exceptions from creating pods, which was the most common prepuller error, but add a try/except block around the entire background thread and report any other uncaught exceptions to Slack. Pause for the image refresh interval after an uncaught exception.


athornton (approved):
LGTM

PR #89, opened 2023-04-14
DM-38554: Report image refresh errors to Slack

rra:
If there is a Slack incoming webhook configured, report errors in the background image refresh thread to Slack.


athornton (approved):
LGTM

PR #92, opened 2023-04-14
DM-38714: Add very basic manual build

rra:
This is currently just the build machinery for a manual and a not-very-great diagram of the internal code layout.


athornton (approved):
Sure.  I guess we will be updating the docs a lot.

PR #94, opened 2023-04-21
DM-38795: Construct a new K8sStorageClient for each request

rra:
Rather than maintaining a global K8sStorageClient, just keep a global kubernetes_asyncio client (to maintain the connection pool) and create new K8sStorageClient objects with each request. This will improve the logging by attaching the per-request logger to the K8sStorageClient when spawning labs, so those log messages will get more context.


athornton (approved):
LGTM.  Might take a little rework for the fileserver, but that's a later bridge.

PR #96, opened 2023-04-25
DM-38795: Fix API documentation

rra:
The URL for the OpenAPI documentation incorrectly had an extra leading slash. Remove the extraneous slash.


rra:
Nope, every slash is significant.

PR #97, opened 2023-04-27
DM-38795: Merge event manager and user map

rra:
The UserMap data structure and the event manager were both tracking information about the user's spawned lab and were both updated at the same points during the spawn process. Eventually, they will both manage data stored in Redis. Merge them into a single concent of a user lab state, handled by LabStateManager, and expose a higher-level API that, instead of making low-level changes to the state, publishes specific events in the lifecycle of the lab.

Move the code to reconcile state from Kubernetes into this new LabStateManager, and run that reconciliation periodically rather than only at startup. This will eventually let us recover from Kubernetes killing pods out from under us, although all of the machinery is not yet present for that and there are some race conditions.


src/jupyterlabcontroller/constants.py
```diff
@@ -3,20 +3,42 @@
...
-"""Default path to injected pod metadata."""
+DROPDOWN_SENTINEL_VALUE = "use_image_from_dropdown"
+"""Used in the lab form for ``image_list`` when ``image_dropdown`` is used."""
 
 IMAGE_REFRESH_INTERVAL = timedelta(minutes=5)
 """How frequently to refresh the list of remote and cached images."""
 
-DROPDOWN_SENTINEL_VALUE = "use_image_from_dropdown"
-"""Used in the lab form for ``image_list`` when ``image_dropdown`` is used."""
+LAB_STATE_REFRESH_INTERVAL = timedelta(minutes=60)
```
athornton, line 32: That's far too long for a default, I think.  5 minutes, maybe?

athornton (approved):
I think I like where you're going with this, although it's gonna take some work to fit the fileserver around.

src/jupyterlabcontroller/constants.py
rra: I'm worried about the overhead here when we have 600 running labs. I'm also changing the code so that when JupyterHub polls the status of a lab, we check Kubernetes directly and detect disappeared lab pods then, so this is just a fallback and cleanup for updating the state of pods we're tracking but that JupyterHub hasn't asked us about, or to fix the state if someone manually deleted the rest of the namespace. For those purposes, I think once an hour should be okay.

PR #98, opened 2023-04-27
tickets/DM-38729: Add fileserver creation

athornton:
This adds a way to create/destroy a fileserver for a given user.


src/jupyterlabcontroller/storage/k8s.py
```diff
@@ -674,3 +682,430 @@ def _standard_metadata(self, name: str) -> V1ObjectMeta:
...
+    right method, and then reuse the guts of creation/retry code?  And
+    then you realize that you could do this to await the creation of
+    an arbitrary Kubernetes object also.
+
+    I am here to warn you, dear reader, that this turns out to be a
+    terrible idea, because it turns errors you could have caught
+    quickly and easily with the type system into runtime errors deep
+    in the FastAPI event loop, with no obvious way to connect the
+    stack trace back to the place you forgot an "await" or misspelled
+    a method name."""
```
rra, line 703: Great docstring, but nit, it should move into the docstring for the class in a `Notes` section rather than free-floating in the middle of the class implementation. :)

src/jupyterlabcontroller/storage/k8s.py
```diff
@@ -5,13 +5,15 @@
 import asyncio
 from base64 import b64encode
 from collections.abc import AsyncGenerator
-from typing import Optional
+from threading import Lock
```
rra, line 8: This is async code so you can't use `threading.Lock`. Best case is that it won't do anything; worst case is that it will deadlock the process. You have to use `asyncio.Lock`.

src/jupyterlabcontroller/storage/k8s.py
```diff
@@ -44,13 +46,19 @@
 
 __all__ = ["K8sStorageClient"]
 
+FILESERVER_LOCK = Lock()
```
rra, line 49: I think the storage layer is the wrong architectural layer to do this synchronization. If the goal here is to prevent racing file server startup in two different requests from the same user, the locking should be at the service layer.

Currently, you have the problem that both the `FileserverManager` and the `K8sStorageClient` are per-request, which is why I moved things into a singleton `LabStateManager`. Either that should become `StateManager` and manage the user file servers as well, or there should be a `FileserverStateManager` that plays a similar role, and then you could put locking there if needed.

src/jupyterlabcontroller/storage/k8s.py
```diff
@@ -674,3 +682,430 @@ def _standard_metadata(self, name: str) -> V1ObjectMeta:
...
+
+    async def create_fileserver(
+        self,
+        username: str,
+        namespace: str,
+        job: V1Job,
+        service: V1Service,
+        gf_ingress: dict[str, Any],
+    ) -> bool:
+        with FILESERVER_LOCK:
```
rra, line 715: I'm not sure I understand what you're trying to do with this locking. What is being gained by creating all of the objects together atomically (except when there's a Kubernetes failure)?

src/jupyterlabcontroller/storage/k8s.py
```diff
@@ -674,3 +682,430 @@ def _standard_metadata(self, name: str) -> V1ObjectMeta:
...
+                    while True:
+                        good = await self.check_fileserver_present(
+                            username, namespace
+                        )
+                        if good:
+                            self._logger.info(
+                                f"Fileserver created for {username}"
+                            )
+                            return True
+                        await asyncio.sleep(interval)
```
rra, line 743: Polling is messy and will make startup unnecessarily slow. I think we should always use event watches instead of polling, since Kubernetes has very nice event watch support.

src/jupyterlabcontroller/storage/k8s.py
```diff
@@ -674,3 +682,430 @@ def _standard_metadata(self, name: str) -> V1ObjectMeta:
...
+    async def _create_fileserver_job(
+        self, username: str, namespace: str, job: V1Job
+    ) -> None:
+        obj_name = f"{username}-fs"
+        self._logger.debug("Creating job", name=obj_name, namespace=namespace)
+        try:
+            await self.batch_api.create_namespaced_job(namespace, job)
+        except ApiException as e:
+            if e.status == 409:
+                # It already exists.  Delete and recreate it
```
rra, line 773: Could you explain more in comment / docstring why this is the right thing to do? I'm not sure I understand.

src/jupyterlabcontroller/storage/k8s.py
```diff
@@ -674,3 +682,430 @@ def _standard_metadata(self, name: str) -> V1ObjectMeta:
...
+
+    async def _create_fileserver_service(
+        self, username: str, namespace: str, spec: V1Service
+    ) -> None:
+        obj_name = f"{username}-fs"
+        try:
+            await self.api.create_namespaced_service(namespace, spec)
+        except ApiException as e:
+            if e.status == 409:
+                # It already exists.  Delete and recreate it
```
rra, line 819: Likewise here, not sure why deleting and recreating on conflict is the right thing to do.

src/jupyterlabcontroller/services/fileserver.py
```diff
@@ -0,0 +1,422 @@
...
+            labels={
+                "argocd.argoproj.io/instance": "fileservers",
+                "lsst.io/category": "fileserver",
+                "lsst.io/user": username,
+            },
+            annotations={
+                "argocd.argoproj.io/compare-options": "IgnoreExtraneous",
+                "argocd.argoproj.io/sync-options": "Prune=false",
+            },
+        )
```
rra, line 212: This duplicates `_standard_metadata` in `K8sStorageClient` except for a different Argo CD project and the `lsst.io` annotations. I think those functions should be merged at an appropriate architectural level. The category and user metadata seems helpful, or at least harmless, for lab objects as well.

src/jupyterlabcontroller/storage/k8s.py
```diff
@@ -674,3 +682,430 @@ def _standard_metadata(self, name: str) -> V1ObjectMeta:
...
+
+        Objects with the name <username>-fs are presumed to be fileserver
+        objects, where <username> can be assumed to be the name of the
+        owning user.
+
+        It returns a dict mapping strings to the value True, indicating those
+        users who currently have fileservers.
+        """
+        observed_state: dict[str, bool] = {}
+        # Create fileserver namespace if necessary.
```
rra, line 933: Stray comment; there's no code here to create it.

src/jupyterlabcontroller/storage/k8s.py
```diff
@@ -674,3 +682,430 @@ def _standard_metadata(self, name: str) -> V1ObjectMeta:
...
+        can do, since we created them and that's the convention we chose.
+
+        A fileserver is working if:
+
+        1) it has a Job with at least one active Pod
+        2) it has a Service (properties not checked)
+        3) it has an Ingress (properties not checked)
+
+        We do not check the GafaelfawrIngress, because the Custom API is
+        clumsy, and the created Ingress is a good proxy.
```
rra, line 980: Well, more to the point, it doesn't matter whether there is a `GafaelfawrIngress`; the fileserver will only work if the `Ingress` exists. If `gafaelfawr-operator` is failing, the fileserver will never come up from a user's perspective.

src/jupyterlabcontroller/storage/k8s.py
```diff
@@ -674,3 +682,430 @@ def _standard_metadata(self, name: str) -> V1ObjectMeta:
...
+
+        If we find a broken fileserver, we delete all its objects.  In
+        steady-state operations, this will happen when a fileserver Pod has
+        seen no operations for its timeout period; the Pod will exit and
+        the reconciliation task will note that the Job has no active Pods
+        and take action to clean it up.
```
rra, line 986: I don't like this architecture for timeouts very much. The pod exits and leaves behind a stranded ingress, which will keep trying to route traffic to a non-existent pod and producing 503 errors until the reconciliation wakes up and cleans up the rest of the objects. If we're doing this, I think we need to have a constant event watch and immediately delete the other objects as soon as the job exits. But that's also a bit awkward to do when the pod is being managed by a job.

I think we may need to have an architectural discussion about this. Having the pod just exit on idle timeout ends up being very awkward in a Kubernetes context. I wonder if it should report its idleness via a web API so that we can collect the idle timeout decision here in the controller and have the controller correctly shut everything down when it's idle (which involves deleting the ingress first before stopping the pod).

src/jupyterlabcontroller/storage/k8s.py
```diff
@@ -674,3 +682,430 @@ def _standard_metadata(self, name: str) -> V1ObjectMeta:
...
+            ) from e
+        if job.status is None or job.status.active < 1:
+            self._logger.info(
+                f"Job {obj_name} has no active pods; "
+                + f"terminating fileserver for {username}"
+            )
+            return False
+        try:
+            self._logger.debug(f"Checking service for {username}")
+            job = await self.api.read_namespaced_service(obj_name, namespace)
```
rra, line 1011: Not sure you really need to do this; I don't see any reason why the `Service` object would get randomly deleted, unlike the `Pod` (which exits) and the `Ingress` (which has to be created by an operator). I suppose it can't hurt, but it may be wasted effort.

src/jupyterlabcontroller/storage/k8s.py
```diff
@@ -674,3 +682,430 @@ def _standard_metadata(self, name: str) -> V1ObjectMeta:
...
+    async def remove_fileserver(self, username: str, namespace: str) -> None:
+        """Remove the set of fileserver objects for a user.  It doesn't
+        return until the objects are no longer present in Kubernetes.
+        """
+        with FILESERVER_LOCK:
+            await self._delete_fileserver_job(username, namespace)
+            await self._delete_fileserver_service(username, namespace)
+            await self._delete_fileserver_gafaelfawringress(
+                username, namespace
+            )
```
rra, line 1048: As mentioned elsewhere, this logic should move to the service layer; it's too much business logic for the storage layer. But when doing that, the ingress should be deleted first, then the service, then the job.

src/jupyterlabcontroller/storage/k8s.py
```diff
@@ -674,3 +682,430 @@ def _standard_metadata(self, name: str) -> V1ObjectMeta:
...
+                        f"Error waiting for {kind} deletion",
+                        e,
+                        namespace=namespace,
+                        name=obj_name,
+                    ) from e
+                self._logger.debug(
+                    f"{kind} still present; waiting {interval}s "
+                    + "then rechecking."
+                )
+                await asyncio.sleep(interval)
```
rra, line 1111: As mentioned elsewhere, we should prefer event watches to polling.

tests/configs/cycle/input/config.yaml
```diff
@@ -6,6 +6,8 @@ lab:
     small:
       cpu: 1.0
       memory: 3Gi
+fileserver:
+  image: docker.io/lsstsqre/worblehat
```
rra, line 10: If we make the file server optional, the specialty configs for testing other things won't need (and shouldn't have) file server configurations.

tests/handlers/fileserver_test.py
```diff
@@ -0,0 +1,66 @@
+"""Tests for user lab routes."""
```
rra, line 1: This is a good start for tests. Testing also needed:
- Behavior when the file server already exists
- Reconciliation when the job exits due to a timeout (if we keep that architecture)
- Timeout where the job never starts
- Timeout where the ingress never gets created
- Should start the file server creation first, and then create the ingress, so that you check that the waiting for ingress creation is done properly

tests/support/kubernetes.py
```diff
@@ -239,7 +251,12 @@ async def create_namespaced_custom_object(
             plural,
             body,
         )
-        assert body["api_version"] == f"{group}/{version}"
+        # Asyncio Kubernetes does not do camelCase to snake_case conversion,
+        # but it eventually probably should.  In the meantime, accomodate it.
+        try:
+            assert body["api_version"] == f"{group}/{version}"
+        except KeyError:
+            assert body["apiVersion"] == f"{group}/{version}"
```
rra, line 259: The mock should exactly match the behavior of the thing it's mocking, so if kubernetes_asyncio doesn't do camel-case conversion, neither should we. In other words, my previous code was just buggy (sorry about that; if it weren't, you would have caught this problem earlier).

tests/support/kubernetes.py
```diff
@@ -468,6 +522,127 @@ async def delete_namespaced_config_map(
...
+            Namespace in which to create the deployment.
+        body
+            Deployment specification.
+
+        Raises
+        ------
+        ApiException
+            Raised with 409 status if the deployment already exists.
+        """
+        self._maybe_error("create_namespaced_pod", namespace, body)
```
rra, line 551: Cut and paste error here.

tests/support/kubernetes.py
```diff
@@ -468,6 +522,127 @@ async def delete_namespaced_config_map(
...
+        name = body.metadata.name
+        await self.create_namespaced_replica_set(
+            namespace,
+            V1ReplicaSet(
+                metadata=V1ObjectMeta(name=name, namespace=namespace)
+            ),
+        )
+        body.status = V1DeploymentStatus(available_replicas=1)
+        self._update_metadata(body, "v1", "Deployment", namespace)
+        self._store_object(namespace, "Deployment", name, body)
```
rra, line 561: You should do these before creating the `ReplicaSet` because that's the order in which things happen in Kubernetes, and you also want to do the error checking done by `_update_metadata` and `_store_object` first.

tests/support/kubernetes.py
```diff
@@ -468,6 +522,127 @@ async def delete_namespaced_config_map(
...
+        V1Status
+            Success status.
+
+        Raises
+        ------
+        ApiException
+            Raised with 404 status if the pod was not found.
+        """
+        await self.delete_namespaced_replica_set(name, namespace)
+        self._maybe_error("delete_namespaced_deployment", name, namespace)
```
rra, line 587: `_maybe_error` must always be first.

tests/support/kubernetes.py
```diff
@@ -580,6 +755,177 @@ async def readline() -> bytes:
         response.content.readline.side_effect = readline
         return response
 
+    # INGRESS API
+
+    async def create_namespaced_ingress(
+        self, namespace: str, body: V1Ingress
+    ) -> None:
+        """Create an ingress object.  This will be given a dummy value
```
rra, line 763: Nit: Second sentence isn't part of the summary and should be a separate paragraph.

tests/support/kubernetes.py
```diff
@@ -580,6 +755,177 @@ async def readline() -> bytes:
...
+        ApiException
+            Raised with 404 status if the ingress was not found.
+        """
+        self._maybe_error("read_namespaced_ingress", name, namespace)
+        return self._get_object(namespace, "Ingress", name)
+
+    # JOB API
+
+    async def create_namespaced_job(self, namespace: str, body: V1Job) -> None:
+        """Create a job object.  This will in turn create a pod object.  That
```
rra, line 847: Nit: Second and subsequent sentences aren't part of the summary and should be separate paragraphs.

tests/support/kubernetes.py
```diff
@@ -580,6 +755,177 @@ async def readline() -> bytes:
...
+        ApiException
+            Raised with 409 status if the object already exists.
+
+        """
+        self._maybe_error("create_namespaced_job", namespace, body)
+        name = body.metadata.name
+        await self.create_namespaced_pod(
+            namespace,
+            V1Pod(
+                metadata=V1ObjectMeta(name=name),
```
rra, line 869: Nit: Formatting will be better if you delete that trailing comma.

tests/support/kubernetes.py
```diff
@@ -580,6 +755,177 @@ async def readline() -> bytes:
...
+            ),
+        )
+        pod = await self.read_namespaced_pod(name, namespace)
+        pod.status.phase = "Running"
+        self._update_metadata(pod, "v1", "Pod", namespace)
+        body.status = V1JobStatus(active=1)
+        self._update_metadata(body, "batch/v1", "Job", namespace)
+        self._store_object(namespace, "Job", name, body)
```
rra, line 877: As above, these should be done first before creating the pod.

tests/support/kubernetes.py
```diff
@@ -580,6 +755,177 @@ async def readline() -> bytes:
...
+        self._maybe_error("create_namespaced_job", namespace, body)
+        name = body.metadata.name
+        await self.create_namespaced_pod(
+            namespace,
+            V1Pod(
+                metadata=V1ObjectMeta(name=name),
+            ),
+        )
+        pod = await self.read_namespaced_pod(name, namespace)
+        pod.status.phase = "Running"
```
rra, line 873: Don't do this -- the newly created pod should honor the default pod phase setting.

tests/support/kubernetes.py
```diff
@@ -871,6 +1218,126 @@ async def read_namespaced_pod_status(
...
+        self._update_metadata(pod, "v1", "Pod", namespace)
+        self._update_metadata(body, "v1", "ReplicaSet", namespace)
+        self._store_object(namespace, "ReplicaSet", name, body)
```
rra, line 1256: As above, this should be done before creating the pod.

tests/support/kubernetes.py
```diff
@@ -871,6 +1218,126 @@ async def read_namespaced_pod_status(
...
+        self._maybe_error("create_namespaced_replica_set", namespace, body)
+        name = body.metadata.name
+        await self.create_namespaced_pod(
+            namespace,
+            V1Pod(
+                metadata=V1ObjectMeta(name=name),
+            ),
+        )
+        pod = await self.read_namespaced_pod(name, namespace)
+        pod.status.phase = "Running"
```
rra, line 1253: As above, don't do this; let the standard pod phase handling happen.

tests/support/kubernetes.py
```diff
@@ -871,6 +1218,126 @@ async def read_namespaced_pod_status(
...
+        V1Status
+            Success status.
+
+        Raises
+        ------
+        ApiException
+            Raised with 404 status if the replica_set was not found.
+        """
+        await self.delete_namespaced_pod(name, namespace)
+        self._maybe_error("delete_namespaced_replica_set", name, namespace)
```
rra, line 1282: `_maybe_error` should be first.

rra (commented):
I can help with the refactoring for the changes I made to state handling and with some of the other larger pieces here. Let's chat about the details next week.

rra:
For the work I've been doing, I also deleted `tests.support.kubernetes` in favor of using the version from Safir and have started adding the required additional methods to the mock provided in Safir. (https://github.com/lsst-sqre/safir/pull/162 is the first.) You'll want to do the same thing, since we don't want to keep two copies of this mock.

athornton:
I have (for ticket hygiene reasons) moved the work to DM-38972.  I'm going to PR that one, but address the comments from this one over there.

src/jupyterlabcontroller/assets/gafaelfawr_ingress_template.txt
```diff
@@ -0,0 +1,29 @@
+apiVersion: gafaelfawr.lsst.io/v1alpha1
```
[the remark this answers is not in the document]
athornton: Deleted

src/jupyterlabcontroller/config.py
```diff
@@ -265,6 +298,7 @@ def _validate_secrets(cls, v: list[LabSecret]) -> list[LabSecret]:
 class Config(BaseSettings):
     safir: SafirConfig
     lab: LabConfig
+    fileserver: FileserverConfig
```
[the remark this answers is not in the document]
athornton: ...I think we are, and we don't want to make it optional.  Might be a Frossie question.  I'll go ask.

...yes, make it opt-in.

src/jupyterlabcontroller/handlers/fileserver.py
```diff
@@ -0,0 +1,92 @@
...
+    themselves have a fileserver and create it if it doesn't exist.
+
+    In all these cases, we provide a redirect to their own fileserver
+    endpoint.
+
+    Finally, if they go to "/files/<someone-else>" and there already
+    is an ingress then they will go there (and we will never see the
+    request), but they won't have a token that lets them authenticate
+    to WebDAV with it, so they can't do anything nefarious.
+    """
```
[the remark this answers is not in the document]
athornton: Well, there's no actual handling for that case.  There's a more-specific ingress that handles the request.  But if I go to /files/rra, and you already have a fileserver running, I'm going to get prompted to authenticate, and I presumably do not have your token.

I will make this not be a docstring.

src/jupyterlabcontroller/handlers/fileserver.py
```diff
@@ -0,0 +1,92 @@
...
+    username = x_auth_request_user
+    gafaelfawr_client = context.factory.create_gafaelfawr_client()
+    user = await gafaelfawr_client.get_user_info(x_auth_request_token)
+    if user.username != username:
+        raise PermissionDeniedError("Permission denied")
```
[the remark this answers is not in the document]
athornton: Noted.

src/jupyterlabcontroller/handlers/fileserver.py
```diff
@@ -0,0 +1,92 @@
...
+user_router = APIRouter(route_class=SlackRouteErrorHandler)
+"""Router to mount into the application."""
+
+__all__ = ["router", "user_router"]
+
+# This router is really at the top level--there's no safir path prefix.
+
+
+@user_router.get(
+    "/files",
```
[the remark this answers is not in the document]
athornton: I think we really want `/files` but making it a configurable seems like a good idea.  From the user's perspective...is it part of Nublado?

src/jupyterlabcontroller/handlers/fileserver.py
```diff
@@ -0,0 +1,92 @@
...
+    context.rebind_logger(user=username)
+    fileserver_manager = context.factory.create_fileserver_manager()
+    result = await fileserver_manager.create_fileserver_if_needed(user)
+    if result:
+        return f"/files/{username}"
+    raise PermissionDeniedError("Permission denied")
+
+
+# The remaining endpoints are for administrative functions and can be
+# tucked under the safir path prefix
```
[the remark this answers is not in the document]
athornton: Given that one is two endpoints and the other is one...I felt like it didn't make sense, but I am open to refactoring it if you think it should be two different modules.

src/jupyterlabcontroller/services/fileserver.py
```diff
@@ -0,0 +1,422 @@
...
+        self._logger = logger
+        self._tasks: Set[asyncio.Task] = set()
+        self._started = False
+
+    async def start(self) -> None:
+        if not await self.k8s_client.check_namespace(self.fs_namespace):
+            self._logger.warning(
+                "No namespace '{self.fs_namespace}'; cannot start"
+            )
+            return
```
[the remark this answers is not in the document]
athornton: Will do.

src/jupyterlabcontroller/services/fileserver.py
```diff
@@ -0,0 +1,422 @@
...
+        self.k8s_client = k8s_client
+        self._tasks: Set[asyncio.Task] = set()
+
+    async def create_fileserver_if_needed(self, user: UserInfo) -> bool:
+        """If the user doesn't have a fileserver, create it.  If the user
+        already has a fileserver, just return.
+
+        This gets called by the handler when a user comes in through the
+        /file ingress.
+        """
```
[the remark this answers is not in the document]
athornton: It's whether there's a running fileserver for the user when it returns, but, sure, None-and-error should work.

src/jupyterlabcontroller/services/fileserver.py
```diff
@@ -0,0 +1,422 @@
...
+            # these settings, but we really don't intend to allow the RSP
+            # to be used as a bulk download service.  Sensibly-sized things
+            # such as notebooks (without enormous rendered output, anyway)
+            # are actually pretty snappy.
+            resources=V1ResourceRequirements(
+                limits={
+                    "cpu": "100m",
+                    "memory": "128M",
+                }
+            ),
```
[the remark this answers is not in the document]
athornton: Will do

src/jupyterlabcontroller/services/fileserver.py
```diff
@@ -0,0 +1,422 @@
...
+        username
+            User name
+
+        Returns
+        -------
+        V1ObjectMeta
+            Kubernetes metadata specification for that user's fileserver
+            objects.
+        """
+        obj_name = f"{username}-fs"
```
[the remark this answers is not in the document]
athornton: I would prefer the username first, since that is in some sense the much more important information when you're debugging.

src/jupyterlabcontroller/services/fileserver.py
```diff
@@ -0,0 +1,422 @@
...
+        # FIXME work out tolerations
+        #
+        podspec = V1PodSpec(
+            containers=[container],
+            restart_policy="Never",
+            security_context=V1PodSecurityContext(
+                run_as_user=user.uid,
+                run_as_group=user.gid,
+                run_as_non_root=True,
+                fs_group=user.gid,
```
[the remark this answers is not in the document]
athornton: OK, I'll remove it.

src/jupyterlabcontroller/services/fileserver.py
```diff
@@ -0,0 +1,422 @@
...
+                    ),
+                    name=vname,
+                )
+            vm = V1VolumeMount(
+                mount_path="/mnt" + storage.container_path,
+                read_only=ro,
+                name=vname,
+            )
+            vols.append(VolumeContainer(volume=vol, volume_mount=vm))
+        return vols
```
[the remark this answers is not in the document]
athornton: It's not entirely duplicated, because WebDAV can only access a single mountpoint per client, hence my aggregation of them all under `/mnt`.  I did try this with a call to the Lab Manager's build volumes and then a loop over the Volume Mounts, but it was less obvious what was happening.  I can go back and reinsert that, though.

src/jupyterlabcontroller/services/fileserver.py
```diff
@@ -0,0 +1,422 @@
...
+                f"{apj}/instance": "fileservers",
+                "lsst.io/category": "fileserver",
+                "lsst.io/user": username,
+            },
+            "annotations": {
+                f"{apj}/compare-options": "IgnoreExtraneous",
+                f"{apj}/sync-options": "Prune=false",
+            },
+        }
+        return md_obj
```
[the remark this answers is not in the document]
athornton: Another thing I tried, and wasn't happy with the result, but can go back to.

src/jupyterlabcontroller/storage/k8s.py
athornton: Will do.

src/jupyterlabcontroller/storage/k8s.py
athornton: Oooh, thank you.

src/jupyterlabcontroller/storage/k8s.py
athornton: I have a FileserverStateManager now, and I can move the locking into it.

src/jupyterlabcontroller/storage/k8s.py
athornton: What I am trying to accomplish is that if a deletion is going on, you can't recreate the user at the same time.  Deletion and creation are both pretty fast, so forcing them through a one-at-a-time bottleneck didn't seem so bad.

src/jupyterlabcontroller/storage/k8s.py
athornton: That sounds like it's gonna be painful but, OK.

src/jupyterlabcontroller/storage/k8s.py
athornton: The fileserver itself has no meaningful state.  If there are somehow leftover bits of a previous fileserver when we try to create a new one, we want to just destroy the previous pieces, and then continue with our object creation.

src/jupyterlabcontroller/services/fileserver.py
athornton: I can do that.  I didn't want to start exporting a bunch more stuff from the storage client, but, OK.

src/jupyterlabcontroller/storage/k8s.py
athornton: Removed

src/jupyterlabcontroller/storage/k8s.py
athornton: Well, adding a user route for fileserver deletion, making sure that the Pod has the appropriate user token, and triggering that (so "Exit without saving" in JupyterLab-extension-land), would work.  It makes Worblehat slightly more complicated (which will about double its complexity), though

src/jupyterlabcontroller/storage/k8s.py
athornton: Will do with regards to the ordering.

tests/configs/cycle/input/config.yaml
athornton: will fix

tests/handlers/fileserver_test.py
athornton: noted; will fix

tests/support/kubernetes.py
athornton: Yup.

tests/support/kubernetes.py
athornton: All the deployment stuff has gone away since I moved to a Safir branch and I'm not using it and the ReplicaSets anymore.

tests/support/kubernetes.py
athornton: Will it, or will black just reinsert it?

tests/support/kubernetes.py
athornton: Same as deployment, the replicaSet mocks no longer exist.

src/jupyterlabcontroller/handlers/fileserver.py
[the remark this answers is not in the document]
athornton: Done.

src/jupyterlabcontroller/handlers/fileserver.py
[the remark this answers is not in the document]
athornton: Operationally, how do I make the path configurable?  Like, obviously I can add it to the config, but in operation I can't get to the config until I'm in the route handler, and the decorator to the route handler needs the path.

src/jupyterlabcontroller/services/fileserver.py
```diff
@@ -0,0 +1,422 @@
...
+        if missing_users:
+            self._logger.info(
+                f"Users {missing_users} have broken fileservers; removing."
+            )
+        for user in missing_users:
+            rmuser_task = asyncio.create_task(
+                self.k8s_client.remove_fileserver(user, namespace)
+            )
+            self._tasks.add(rmuser_task)
+            rmuser_task.add_done_callback(self._tasks.discard)
```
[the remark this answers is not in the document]
athornton: Foreground removals it is.

Will think through the concurrency later.

src/jupyterlabcontroller/services/fileserver.py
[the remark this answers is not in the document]
athornton: Have done.

src/jupyterlabcontroller/services/fileserver.py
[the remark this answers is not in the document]
athornton: Removed.

src/jupyterlabcontroller/services/fileserver.py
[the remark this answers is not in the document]
athornton: Moved into LabBuilder, which eventually will need a more generic name, but let's get the DM-38795 work merged first.

src/jupyterlabcontroller/storage/k8s.py
athornton: Done

src/jupyterlabcontroller/storage/k8s.py
athornton: Have done.

src/jupyterlabcontroller/services/fileserver.py
athornton: I've done this although I feel like ultimately the metadata builder belongs in the Builder object, not the storage client.

src/jupyterlabcontroller/config.py
```diff
@@ -256,6 +256,39 @@ def _validate_secrets(cls, v: list[LabSecret]) -> list[LabSecret]:
...
+        example="Always",
+        title="Pull policy for the fileserver image",
+    )
+    timeout: int = Field(
+        3600, title="Inactivity timeout for the fileserver container (seconds)"
+    )
+    namespace: str = Field(
+        FILESERVER_NAMESPACE,
+        title="Namespace for user fileservers",
+    )
```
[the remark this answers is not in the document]
athornton: We had a Slack discussion; I think I'm reasonably happy with what I have now.

src/jupyterlabcontroller/config.py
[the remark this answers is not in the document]
athornton: Turns out that making it actually Optional pushes a bunch of checking elsewhere.  What I eventually went with is to make the default configuration broken, in the sense that neither image nor namespace are set, but also add an enabled flag to the fileserver config (default: False), and only do fileservery things if enabled is set.  It uses a root validator on the fileserver config to ensure that image and namespace have values if enabled is turned on.

src/jupyterlabcontroller/handlers/fileserver.py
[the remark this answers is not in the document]
athornton: I can set a prefix, which could be the empty string.  Done.

src/jupyterlabcontroller/services/fileserver.py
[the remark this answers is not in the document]
athornton: Well, this turned into a can of worms, but, working now.

src/jupyterlabcontroller/services/fileserver.py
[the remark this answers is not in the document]
athornton: Done.

tests/configs/cycle/input/config.yaml
athornton: Done

tests/support/kubernetes.py
athornton: removed

tests/support/kubernetes.py
athornton: Deployment/Replicaset have gone away for now, because I don't need them.

tests/support/kubernetes.py
athornton: done, anyway.

tests/support/kubernetes.py
athornton: done

tests/support/kubernetes.py
athornton: Replicaset/Deployment are gone

tests/support/kubernetes.py
athornton: Job does it; ReplicaSet/Deployment are deferred until we need them

tests/support/kubernetes.py
athornton: removed (there's no deployment anymore)

tests/support/kubernetes.py
athornton: Done

tests/support/kubernetes.py
athornton: Done

src/jupyterlabcontroller/storage/k8s.py
athornton: I'd really like to leave Worblehat trivially simple.  Doing a watch on the job seems pretty easy though.

src/jupyterlabcontroller/services/fileserver.py
```diff
@@ -0,0 +1,422 @@
...
+            tsk.cancel()
+
+    async def reconcile_user_map(self) -> None:
+        self._logger.debug("Reconciling fileserver user map")
+        namespace = self.fs_namespace
+        if not await self.k8s_client.check_namespace(namespace):
+            self._logger.warning(
+                f"No fileserver namespace '{namespace}'; no fileserver users"
+            )
+            return
```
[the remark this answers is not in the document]
athornton: Done.

src/jupyterlabcontroller/storage/k8s.py
athornton: Amended.

src/jupyterlabcontroller/storage/k8s.py
athornton: Taken out.

src/jupyterlabcontroller/storage/k8s.py
athornton: Done (with respect to the ordering).  Service layer migration deferred till later.

src/jupyterlabcontroller/storage/k8s.py
athornton: ...kinda.  Without a "parent" pointer, I can at least make the StateManager's lock available to its child processes, which should do the trick.

src/jupyterlabcontroller/services/fileserver.py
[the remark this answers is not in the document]
athornton: I've added locking.  Not sure it's in quite the right place, but basically there's a shared lock that protects access to the user map, and is asserted around fileserver creation and deletion.  I don't expect this to be massively parallel, because this is just the interface a user uses to start a fileserver, and then they ignore the endpoint until they let the server time out and want to do WebDAVy things again...at which point they will notice their client is broken, and go to /files again.  I doubt if we ever even get to more-than-one-a-second in actual use.  (Sure, we'll have lots of people going through created ingresses to their fileservers, but that's the reason Worblehat should be very lightweight.)

src/jupyterlabcontroller/services/fileserver.py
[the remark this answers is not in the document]
athornton: Did, and I think the resulting labels and annotations look better.

src/jupyterlabcontroller/storage/k8s.py
athornton: We can argue about where exactly the locks belong, but I have a system now that feels a little better.

src/jupyterlabcontroller/storage/k8s.py
athornton: Comments added.

src/jupyterlabcontroller/storage/k8s.py
athornton: comments added.

src/jupyterlabcontroller/storage/k8s.py
athornton: I've reworked the lock a bit but I still don't know if it's right.

src/jupyterlabcontroller/services/fileserver.py
```diff
@@ -0,0 +1,422 @@
...
+        gf_ingress = self.build_fileserver_ingress(user.username)
+        service = self.build_fileserver_service(user.username)
+        # We put this all in k8s_client so it can manage its own lock
+        return await self.k8s_client.create_fileserver(
+            username=username,
+            namespace=namespace,
+            job=job,
+            service=service,
+            gf_ingress=gf_ingress,
+        )
```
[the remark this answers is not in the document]
athornton: moved

src/jupyterlabcontroller/handlers/fileserver.py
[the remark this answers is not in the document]
athornton: Resolving, but if you care a lot I'll change it.

src/jupyterlabcontroller/services/fileserver.py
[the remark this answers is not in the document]
athornton: Done (and this one is in its own commit for easy removal if we want, but I like it better this way)

src/jupyterlabcontroller/storage/k8s.py
athornton: I've done this for the Pod.  Haven't yet extended watches to the non-Pod objects.  It wasn't that painful.

src/jupyterlabcontroller/storage/k8s.py
athornton: Which I've done, actually a watch on the dependent Pod, and it works instantly.

src/jupyterlabcontroller/storage/k8s.py
athornton: Going to resolve this.  More watch work needed, but that will happen on DM-38972

tests/handlers/fileserver_test.py
athornton: Will copy to DM-38972

src/jupyterlabcontroller/storage/k8s.py
athornton: Non-Pod event watches will be address on DM-38972

athornton:
@rra if that looks like we are up-to-date with respect to DM-38972 and that I have adequately captured the remaining work, please close this PR (without merging).

rra:
Replaced with #103.

PR #99, opened 2023-04-28
DM-38795: Overhaul task tracking for lab spawns

rra:
Previously, when someone requested that a lab be spawned, this created a background task recorded in the `LabManager` that created the Kubernetes objects and then waited for the pod to finish starting. The drawback of this approach is that nothing ever reaped this background task to properly collect its return status or exception, resulting in Python errors about abandoned tasks.

Now that `LabManager` exists, hand over management of spawner tasks to it. Provide it with a callback that creates the Kubernetes objects, and move waiting for the lab to start into `LabStateManager`. These threads are now properly reaped by a long-running reaper background task. Move the exception handling and reporting for the spawner task into `LabStateManager` as well.

In addition to the benefit of better task tracking and management, this also lays the groundwork for spawning a monitoring task during startup if we detect a lab spawn in progress. It also allows making more of the event posting and state update machinery internal to `LabStateManager`, since it now knows when lab spawns complete and doesn't have to be told by other code.

Internally, collect all of the state information for a user's lab into a single data structure, which avoids accidental state inconsistencies. Use that data structure to skip reconciliation with Kubernetes when we know that actions are in progress on that lab, since otherwise we may race with ourselves. As a side effect, this means we no longer retain deletion events for a lab after the lab has been deleted, but on reflection I think this is correct REST semantics and we have no obvious use for those events.

### DIFF
--- a/src/jupyterlabcontroller/handlers/labs.py
+++ b/src/jupyterlabcontroller/handlers/labs.py
@@ -41,7 +41,7 @@ async def get_lab_users(
     },
     summary="Status of user's lab",
 )
-async def get_state(
+async def get_lab_state(
     username: str,
     context: RequestContext = Depends(context_dependency),
 ) -> UserLabState:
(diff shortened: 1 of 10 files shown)


athornton (approved):
This does seem like an improvement.

PR #100, opened 2023-05-01
DM-38795: Merge waiting for pod with event monitoring

rra:
Merge the method to wait for a pod to start with the method that monitors events for that pod and eliminate the extra background task for event injection. Use that same method for the prepuller.

Add a new configurable spawn timeout instead of hard-coding it, and pass that timeout into the Kubernetes watch, using a shorter watch for the network connection timeout.

This removes the suppression of duplicate messages, since it's a bit more awkward to do in this format. We can add it back in, but I wanted to try running this version first and see to what degree duplicate messages are an issue.

### DIFF
--- a/src/jupyterlabcontroller/config.py
+++ b/src/jupyterlabcontroller/config.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import os
+from datetime import timedelta
 from enum import Enum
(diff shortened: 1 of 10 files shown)


athornton (approved):
And just when I'd gotten the fileserver manager looking kinda like the lab manager...

PR #101, opened 2023-05-01
DM-38795: Redo how labs are deleted

rra:
Use the same division of labor between LabManager and LabStateManager as for lab creation, which lets us eliminate more public APIs from LabStateManager for the individual events involved in deletion. As with lab spawning, LabStateManager now handles the exceptions, although lab deletion is done in the foreground since JupyterHub does it in the foreground anyway and therefore no purpose is served by adding background deletion.

### DIFF
--- a/src/jupyterlabcontroller/models/v1/event.py
+++ b/src/jupyterlabcontroller/models/v1/event.py
@@ -26,7 +26,7 @@ class Event(BaseModel):
     type: EventType = Field(..., title="Type of the event")
     message: str = Field(..., title="Event message")
     progress: Optional[int] = Field(
-        None, title="Progress percentage", lt=100, gt=0
+        None, title="Progress percentage", le=100, gt=0
     )
 
     @property
--- a/src/jupyterlabcontroller/services/state.py
+++ b/src/jupyterlabcontroller/services/state.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import asyncio
-from collections.abc import AsyncIterator, Awaitable, Callable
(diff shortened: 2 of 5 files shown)


src/jupyterlabcontroller/models/v1/event.py
```diff
@@ -26,7 +26,7 @@ class Event(BaseModel):
     type: EventType = Field(..., title="Type of the event")
     message: str = Field(..., title="Event message")
     progress: Optional[int] = Field(
-        None, title="Progress percentage", lt=100, gt=0
```
athornton, line 29: Did we ever really define whether each endpoint was open or closed?

src/jupyterlabcontroller/services/state.py
```diff
@@ -342,7 +275,83 @@ async def start_spawn(
         task = asyncio.create_task(
             self._spawn(username, spawner, start_progress, end_progress)
         )
-        self._labs[username].spawner = task
+        self._labs[username].task = task
```
athornton, line 278: I definitely prefer this to publish_start_deletion()

athornton (approved):
LGTM

src/jupyterlabcontroller/models/v1/event.py
rra: I think it's closed, but JupyterHub should get 100% for itself for spawning since it has to wait for the lab service to start. It doesn't really matter for deletion, since nothing looks at those events except us. Using 100% for finishing the deletion felt right.

PR #102, opened 2023-05-01
DM-38795: Get the actual state of the lab from Kubernetes

rra:
When polled by JupyterHub, return the actual state of the lab rather than trusting that in-memory data is correct. This supplements the background reconciliation job. Unlike a reconciliation, this doesn't fix data, just moves things into a failed state if the pod should be running and isn't.


athornton (approved):
Seems good.  I independently did something similar for the fileserver controller.

PR #104, opened 2023-05-01
DM-38795: Delete lab on spawn if in failed state

rra:
If the user already has a lab but that lab is in a failed state, allow them to spawn again and delete the existing lab before creating a new one.

### DIFF
--- a/src/jupyterlabcontroller/services/lab.py
+++ b/src/jupyterlabcontroller/services/lab.py
@@ -43,6 +43,7 @@ from ..models.domain.lab import LabVolumeContainer
 from ..models.domain.rspimage import RSPImage
 from ..models.v1.lab import (
     LabSpecification,
+    LabStatus,
     UserInfo,
     UserLabState,
     UserResourceQuantum,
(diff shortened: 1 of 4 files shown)


src/jupyterlabcontroller/services/lab.py
```diff
@@ -120,33 +121,46 @@ async def create_lab(
...
-        # handed off to LabStateManager here.
+        # raises an exception if the lab already exists and is not in a failed
+        # state.
         #
-        # FIXME: Handle labs in failed state, which should be removed before
-        # starting to create a new lab.
-        self._logger.info("Creating new lab")
+        # A LabManager is per-request, so the management of the background
+        # task that does the lab spawning (which outlasts the request that
+        # kicks it off) is handed off to LabStateManager here.  starting to
```
athornton, line 135: "starting to create a new lab" probably should be deleted?

athornton (approved):
Looks like it should work.

I know I rejected the idea of doing a pair of generic create/destroy wrappers, which would delete-objects-and-try-again on 409 on creation and just-say-you-succeeded on 404 on deletion, but maybe we *do* want that functionality?

rra:
I don't think recreating on 409 is the right generic behavior. It feels like it's going to hide bugs; particularly for lab spawning, if objects exist that we don't expect, that's not a normal and expected problem, that's a startling development that we should investigate.

PR #105, opened 2023-05-02
DM-38795: Fix URL of running lab

rra:
The Jupyter lab takes its base URL from JUPYTERHUB_SERVICE_PREFIX in the lab environment, which is always set by JupyterHub. Although there is a redirect from the top-level URL, ensure that the internal URL we return is fully correct so that JupyterHub doesn't complain about it and change it during startup.

Make JUPYTERHUB_SERVICE_PREFIX required in the lab environment so that we don't have to do error handling, reporting, or fallback. This is unconditionally sent by JupyterHub, so we can make it part of the API contract.


athornton (approved):
This all looks good, but you should test it in the wild before merging, if you haven't already.  Specifically, I remember that we had to use the internal URL in order for Lab-Hub communication to work, but that was a very long time ago.  Go ahead and use data-dev; I don't think I'm gonna work on the fileserver much today.

I wonder if this will, as a side effect, fix the logout menu item.

rra:
Verified on data-dev and this gets rid of the error message on startup of the hub.

PR #108, opened 2023-05-04
DM-38795: Switch to the new Safir Kubernetes mock

rra:
Temporarily switch to using Safir from a branch to get an unreleased version of the Kubernetes mock. (We will switch back to a released version of Safir before the next release.) This mock is somewhat more sophisticated about resource version tracking, which required some minor fixes to the test suite and to pod event watching.

Use a shorter timeout for spawns for the test suite so that problems with state tracking don't cause it to apparently hang forever.


athornton (approved):
Sure.  I'll rebase my safir branch from your safir branch (although maybe I already have since I did the -queue one, so perhaps all your changes are in there?) to add support for the fileserver objects.

PR #109, opened 2023-05-04
DM-38795: Fix lab state reconciliation

rra:
When the lab controller was restarted, its internal data structures were only partly refreshed from the running labs in Kubernetes. A lot of data was missing or incorrect, such as all of the environment variables. This ran afoul of the new validator to ensure that JUPYTERHUB_SERVICE_PREFIX is set.

Fully restore internal state by also reading the ConfigMap and ResourceQuota objects, and add the data that isn't otherwise easily discernable from lab resources (full name and group name information, which only otherwise exist in the templated NSS files) to annotations on the pod.

This introduces a new LabBuilder object, which eventually will hold all the code to construct Kubernetes objects for a user's lab environment (moved from LabManager, which will then only be about managing lab lifecycle). For the time being, this only holds the methods that need to be shared between LabManager and LabStateManager.

### DIFF
--- a/src/jupyterlabcontroller/models/v1/lab.py
+++ b/src/jupyterlabcontroller/models/v1/lab.py
@@ -1,9 +1,10 @@
-"""Models for jupyterlab-controller."""
+"""API-visible models for user lab environments."""
+
+from __future__ import annotations
 
 from enum import Enum
 from typing import Any, Optional, Self
 
-from kubernetes_asyncio.client.models import V1Pod
 from pydantic import BaseModel, Field, root_validator, validator
 
 from ...constants import (
@@ -11,7 +12,6 @@ from ...constants import (
     GROUPNAME_REGEX,
     USERNAME_REGEX,
 )
-from ...util import str_to_bool
 
 __all__ = [
     "ImageClass",
@@ -31,9 +31,11 @@ __all__ = [
 
 
 class LabSize(str, Enum):
-    """Allowable names for pod sizes."""
+    """Allowable names for pod sizes.
 
-    # https://www.d20srd.org/srd/combat/movementPositionAndDistance.htm#bigandLittleCreaturesInCombat
+    Taken from `d20 creature sizes`_.
+    """
+
     FINE = "fine"
     DIMINUTIVE = "diminutive"
     TINY = "tiny"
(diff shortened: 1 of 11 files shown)


src/jupyterlabcontroller/models/v1/lab.py
```diff
@@ -31,9 +31,12 @@
 
 
 class LabSize(str, Enum):
-    """Allowable names for pod sizes."""
+    """Allowable names for pod sizes.
```
athornton, line 34: I prefer quoting the actual SRD document rather than a derivative work, even if the derivative work does have a much more descriptive URL.

rra: Well, more to the point it's a much shorter URL so I can move it into a docstring without flake8 whining at me (otherwise, yes, I'm on board).

src/jupyterlabcontroller/services/builder.py
```diff
@@ -0,0 +1,108 @@
+"""Construction of Kubernetes objects for user lab environments."""
```
athornton, line 1: Nifty approach.

src/jupyterlabcontroller/services/state.py
```diff
@@ -475,6 +493,151 @@ async def _clear_events(self, lab: UserLab) -> None:
...
+        # Gather the necessary information from the pod. If anything is
+        # missing or in an unexpected format, log an error and return None.
+        try:
+            lab_container = None
+            for container in pod.spec.containers:
+                if container.name == "notebook":
+                    lab_container = container
+                    break
+            if not lab_container:
+                raise ValueError("No container named notebook")
```
athornton, line 554: Maybe throw quotes around notebook?  Maybe not.

src/jupyterlabcontroller/services/state.py
```diff
@@ -475,6 +493,151 @@ async def _clear_events(self, lab: UserLab) -> None:
...
+                pod=PodState.PRESENT,
+                internal_url=internal_url,
+                resources=resources,
+            )
+        except Exception as e:
+            error = f"{type(e).__name__}: {str(e)}"
+            logger.error(f"Invalid lab environment for {username}: {error}")
+            return None
+
+    def _recreate_groups(self, pod: V1Pod) -> list[UserGroup]:
```
athornton, line 591: Cheaty but nice.  I'll do the same in fileserver.

src/jupyterlabcontroller/util.py
```diff
@@ -3,11 +3,3 @@
 
 def deslashify(data: str) -> str:
     return data.replace("/", "_._")
-
```
athornton, line 6: We really don't use str_to_bool anywhere anymore?  Surprising.

athornton (approved):
Couple of minor nits, but generally looks very good.

src/jupyterlabcontroller/models/v1/lab.py
rra: I moved the link to the rST epilogue which avoids the line length issue.

PR #110, opened 2023-05-04
DM-38795: Handle conflicts on names of prepuller pods

rra:
When creating a prepuller pod, if a pod already exists with the same name, remove it first. We may sometimes strand prepuller pods when the lab controller is killed while prepulling (assuming Kubernetes doesn't remove the pods due to their ownership metadata), and it's always safe to recreate those pods.

Also remove prepuller pods in any terminal state, not just success, to reduce the chances we leave stranded pods behind.


athornton (approved):
LGTM.  I originally had it only removing succeeded pods so I could more easily what had gone wrong with the prepuller pods, but now that those are pretty reliably working, we don't need that feature anymore.

PR #111, opened 2023-05-04
DM-38795: Handle pending labs found in reconciliation

rra:
If during lab reconciliation we discover a lab in pending status, we need to monitor it so that we see when it starts, just as if it were a lab we created. Otherwise, labs that were pending when the lab controller restarts will stay pending forever and never show as ready.


athornton (approved):
LGTM.  Thing-lifecycle-management is hard.

PR #112, opened 2023-05-05
DM-38795: Handle timeouts consistently

rra:
Use TimeoutError for all timeouts instead of our own separate exception. Add timeout handling to lab spawning and deletion, and test that timeout handling works properly.


athornton (approved):
LGTM, but I added WaitingForObjectError back in to my own stuff, because the point wasn't that the timeout failed but an expected object didn't appear.

PR #114, opened 2023-05-08
DM-38795: Minor improvements to logging and error handling

rra:
- Retire `models.k8s.Secret` in favor of just using `V1Secret`, since this was internal to the storage layer
- Add more details to log messages during spawn by using a bound logger
- Fix the timeout for watching for pod spawn; `_request_timeout` is apparently the total elapsed time
- Remove some extra logging when getting the pod phase that was cluttering up debug logs
- Log the new pod phase when it changes
- Replace `MissingSecretError` with `MissingObjectError` and make it Slack-reportable
- Avoid putting the text of a Kubernetes error in the Slack message since we add it later as an block
- Use `asyncio.timeout` instead of `asyncio.wait_for` based on upstream recommendations


athornton (approved):
LGTM.  The generic MissingObjectError is what I use for reporting missing namespaces if your preflight check for fileserver fails (and I think I have one other use case that might really be a timeout) so that seems good.  Just using the K8s models is better, I agree.

PR #115, opened 2023-05-08
DM-38795: Use a pod watch to wait for pods to start

rra:
Unfortunately, Kubernetes doesn't create an event when a pod changes phase to running, so the approach of only watching Kubernetes namespace events and checking the pod phase when an event is received doesn't work. This required going back to watching the pod phase and the namespace events in separate tasks. The task management of the event watch task is now done by the pod watch task instead of by `LabManager`.

### DIFF
--- a/src/jupyterlabcontroller/services/state.py
+++ b/src/jupyterlabcontroller/services/state.py
@@ -511,6 +511,49 @@ class LabStateManager:
         for trigger in triggers:
             trigger.set()
 
+    async def _reflect_events(
+        self,
+        *,
+        username: str,
+        pod: str,
+        namespace: str,
+        start_progress: int,
+        end_progress: int,
+    ) -> None:
+        """Monitor Kubernetes events for a pod.
+
+        Translate these into our internal event structure and add them to the
+        event list for this user. Intented to be run as a task and cancelled
+        once the pod spawn completes or times out.
+
+        Parameters
+        ----------
+        username
+            Username for which we're spawning a lab.
+        pod
+            Name of the pod.
+        namespace
+            Namespace of the pod.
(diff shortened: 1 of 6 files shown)


src/jupyterlabcontroller/services/state.py
```diff
@@ -511,6 +511,49 @@ async def _clear_events(self, lab: UserLab) -> None:
         for trigger in triggers:
             trigger.set()
 
+    async def _reflect_events(
```
athornton, line 514: I mean, I really do hate it that we've clawed our way back to the KubeSpawner reflector model, but I guess maybe they really did know what they were doing the first time.

src/jupyterlabcontroller/storage/k8s.py
```diff
@@ -179,11 +179,92 @@ async def remove_completed_pod(
 
     async def wait_for_pod(
         self, pod_name: str, namespace: str
-    ) -> AsyncIterator[KubernetesPodEvent]:
+    ) -> KubernetesPodPhase | None:
```
athornton, line 182: This is the bit that I think we are going to want to replace with the more generic model at https://github.com/lsst-sqre/jupyterlab-controller/blob/59c379859a895c9416b6f6ed95d7ef5a978b804c/src/jupyterlabcontroller/storage/k8s.py#L1526, where what we care about is a transition to another state, and then we wrap that in different thin wrappers specifying expected-initial and desired-terminal state.

athornton (approved):
Looks good.  I have an idea for refactoring the inner watch loop so we don't have to duplicate it all over the places for watching for things other than startup or for watching things other than pods.

PR #117, opened 2023-05-09
DM-38425: Set K8S_NODE_NAME for now

rra:
lsst.rsp already looks for the environment variable K8S_NODE_NAME, so while we want to avoid confusing abbreviations, we need to set both that and KUBERNETES_NODE_NAME for now until lsst.rsp has been updated and rolled out in a new weekly and recommended.


athornton (approved):
LGTM

PR #121, opened 2023-05-16
DM-39232: Add support for persistent volume claims

rra:
USDF requires creation of PVCs for each Jupyter lab pods to mount the underlying storage for home space and project space. Add support for PVCs as a source for mounted lab volumes and, in that case, create a PVC with the given parameters. The supported configuration parameters are a subset of the PVC model that is sufficient for the current USDF requirements.

While making this change, refactor the configuration syntax for lab volumes to separate the types of supported volumes into different Pydantic models so that Pydantic can do validation for us.


tests/configs/standard/input/config.yaml
```diff
@@ -154,17 +156,26 @@ lab:
...
+        path: /share1/project
     - containerPath: /scratch
-      serverPath: /share1/scratch
-      server: 10.13.105.122
       mode: rw
+      source:
+        type: persistentVolumeClaim
+        storageClassName: sdf-home
+        accessModes:
+          - ReadWriteMany
```
athornton, line 175: We should maybe see if we can consolidate these and the old "ro"/"rw" distinction, but it's not urgent.

rra: Hm, or maybe change it so that it exactly matches what Kubernetes does, where there's a boolean `read_only` flag in some places but PVCs get this list of access patterns.

PR #122, opened 2023-05-17
DM-39249: Fix JavaScript for selecting images from dropdown

rra:
In order to move the selection to the dropdown when an image was selected from the dropdown, we were using the onclick attribute of the <select> element. This appears to not work correctly with Safari. Change it to onchange, which hopefully will.

See https://stackoverflow.com/questions/1769374/


athornton (approved):
Here's hoping.

PR #124, opened 2023-05-18
DM-39249: Decrease grace period for lab shutdown

rra:
Follow kubespawner and kill the lab with a grace period of 1s before deleting the namespace. The lab apparently doesn't do anything useful with its grace period, and this will take the user out of the lab faster and avoid the pop-up error messages from the lab being removed from its service.


athornton (approved):
Looks good; let's see if it helps.

PR #126, opened 2023-05-19
DM-39232: Add support for subPath in volume mounts

rra:
USDF dev mounts a volume in one place and a subpath of the same volume in a different place. Add support for specifying a subPath in the volume list.

This is not entirely correct since we still generate two PVCs for the same storage, which I think will work at the USDF because everything is ReadWriteMany, but which would fail with ReadWriteOnce. However, the more correct solution requires a more intrusive configuration change to support mounting the same volume in multiple places, which I'm deferring for later work.


athornton (approved):
This is getting ugly.  I hope each new environment that comes online doesn't need its own special-snowflake volume stuff.

PR #127, opened 2023-05-22
[neophile] Update dependencies

sqrbot:
- Update frozen Python dependencies


rra:
Fixing this separately in conjunction with moving to the new Safir.

PR #131, opened 2023-05-25
DM-39400: Improve Kubernetes exception reporting

rra:
Move the Kubernetes object from a Slack field to a block, since it's too long for a field. Remove the brackets around the kind, since it seems reasonably clear without them. Add the kind to the exception summary as well, and handle a few more formatting edge cases.

KubernetesError was used in a couple of places that weren't Kubernetes API errors. Switch them to either MissingObjectError or a new DuplicateObjectError, and also improve the Slack formatting of those exceptions.

Reorder parameters to follow a more natural naming order of kind, namespace, name (from least specific to most specific), except for the methods that take a name and a namespace, where we keep the annoying Kubernetes API order of name, namespace to avoid confusion.

### DIFF
--- a/src/jupyterlabcontroller/exceptions.py
+++ b/src/jupyterlabcontroller/exceptions.py
@@ -14,13 +14,16 @@ from safir.slack.blockkit import (
     SlackCodeBlock,
     SlackException,
     SlackMessage,
+    SlackTextBlock,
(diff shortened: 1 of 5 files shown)


athornton (approved):
This does look like it regularizes things.

PR #133, opened 2023-05-26
DM-39400: Separate Gafaelfawr models

rra:
The change to include the Gafaelfawr token in the UserInfo model meant that we were leaking the token to API callers asking for the lab status. Solve this problem properly by separating the Gafaelfawr model from the UserInfo model that's part of the lab API, and by creating two Gafaelfawr models, one with a token and one without.

Fix up the typing to expect the model needed at each point in the lab controller code. While doing this, remove the now-unnecessary separate token argument to many of the lab construction methods, allowing additional simplification.

Take this opportunity to clean up the lab status API a bit, since it was getting cluttered by collapsing multiple different types of data into one flat object. Move the user metadata into a user key, and return only the user data of interest to the status of the lab by omitting the user's full quota inforamtion. Instead, add the quota enforced on the user's namespace (as distinct from the resource restrictions on the lab) to the top-level UserInfo model.


rra:
Incidentally, "quantity" is what Kubernetes officially seems to call the thing we've been calling a "quantum," so I switched to the Kubernetes terminology.

athornton (approved):
This all seems reasonable to me.

PR #137, opened 2023-05-26
DM-39400: Handle groups without GIDs

rra:
Gafaelfawr allows the user to be a member of a group with no GID, but this should be ignored for Nublado controller purposes since we can't do anything useful with such groups. Filter them out when constructing the lab and when constructing the user metadata portion of the lab status.

Take advantage of the opportunity to refactor the method that generates the NSS ConfigMap.

### DIFF
--- a/src/jupyterlabcontroller/services/lab.py
+++ b/src/jupyterlabcontroller/services/lab.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import json
 import re
-from copy import copy, deepcopy
+from copy import copy
 from functools import partial
 from pathlib import Path
 from typing import Optional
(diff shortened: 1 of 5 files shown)


src/jupyterlabcontroller/services/lab.py
```diff
@@ -313,30 +315,53 @@ async def create_nss(self, user: GafaelfawrUserInfo) -> None:
...
+        -------
+        dict of str to str
+            Contents of the ``ConfigMap``.
+        """
+        etc_passwd = self.lab_config.files["/etc/passwd"].contents
+        etc_group = self.lab_config.files["/etc/group"].contents
+
+        # Construct the user's /etc/passwd entry. Different sites use
+        # different schemes for constructing the home directory path.
+        match self.lab_config.homedir_schema:
```
athornton, line 340: Match!  ❤️

athornton (approved):
LGTM

PR #141, opened 2023-06-07
Update plan.rst

frossie:


rra (approved):
This whole item is done so you can just delete the whole thing and the section if you want.

rra:
Fixing this in a separate PR to update the current to-do list.

PR #147, opened 2023-07-25
tickets/DM-40183: fix incorrect homedir creation at USDF

athornton:
At USDF, despite our homedir_schema setting, we're creating /home/<username>

I believe this fixes the issue, although I'm not going to merge until I've done some testing at USDF-dev.

The cause appears to be 1) not setting run_as_user in the pod spec, so despite the fact that you end up as the right user with the right groups, it starts with enough privilege to create the directory), which 2) is set as working_dir in the pod spec.

So I've corrected these two things.

In an idea world we would have a test for the pod spec being correct when a non-default homedir_schema is in effect, but we didn't make it easy to unit test pod spec creation, which (necessarily) relies on resolved RSPImages, and therefore needs a bunch of new helper methods defined, or an end-to-end handler test (which might actually be easier) checking on the created pod.  I wanted to get USDF out of its current pickle quickly, and implementing the test is going to take some thought and time.

### DIFF
--- a/src/jupyterlabcontroller/services/lab.py
+++ b/src/jupyterlabcontroller/services/lab.py
@@ -877,9 +877,6 @@ class LabManager:
             image_pull_secrets=pull_secrets,
             restart_policy="OnFailure",
             security_context=V1PodSecurityContext(
-                run_as_non_root=True,
-                run_as_user=user.uid,
-                run_as_group=user.gid,
                 supplemental_groups=[x.id for x in user.groups if x.id],
             ),
             volumes=volumes,
--- a/src/jupyterlabcontroller/services/lab.py
+++ b/src/jupyterlabcontroller/services/lab.py
@@ -831,7 +831,6 @@ class LabManager:
             env.append(env_var)
 
         # Specification for the user's container.
-        homedir = self._get_homedir(user)
         container = V1Container(
             name="notebook",
             args=["/opt/lsst/software/jupyterlab/runlab.sh"],
@@ -862,7 +861,7 @@ class LabManager:
                 run_as_group=user.gid,
             ),
             volume_mounts=mounts,
-            working_dir=homedir,
+            working_dir=self._get_homedir(user),
         )
 
         # Build the pod specification itself.
--- a/src/jupyterlabcontroller/services/lab.py
+++ b/src/jupyterlabcontroller/services/lab.py
@@ -878,7 +878,8 @@ class LabManager:
             restart_policy="OnFailure",
             security_context=V1PodSecurityContext(
                 run_as_non_root=True,
-                fs_group=user.gid,
+                run_as_user=user.uid,
+                run_as_group=user.gid,
                 supplemental_groups=[x.id for x in user.groups if x.id],
             ),
             volumes=volumes,
--- a/src/jupyterlabcontroller/services/lab.py
+++ b/src/jupyterlabcontroller/services/lab.py
@@ -146,6 +146,13 @@ class LabManager:
             end_progress=75,
         )
 
+    def _get_homedir(self, user: GafaelfawrUserInfo) -> str:
+        match self.lab_config.homedir_schema:
+            case UserHomeDirectorySchema.USERNAME:
+                return f"/home/{user.username}"
+            case UserHomeDirectorySchema.INITIAL_THEN_USERNAME:
+                return f"/home/{user.username[0]}/{user.username}"
+
     async def _spawn_lab(
         self,
         *,
@@ -337,11 +344,7 @@ class LabManager:
 
         # Construct the user's /etc/passwd entry. Different sites use
         # different schemes for constructing the home directory path.
-        match self.lab_config.homedir_schema:
-            case UserHomeDirectorySchema.USERNAME:
-                homedir = f"/home/{user.username}"
-            case UserHomeDirectorySchema.INITIAL_THEN_USERNAME:
-                homedir = f"/home/{user.username[0]}/{user.username}"
+        homedir = self._get_homedir(user)
         etc_passwd += (
             f"{user.username}:x:{user.uid}:{user.gid}:"
             f"{user.name}:{homedir}:/bin/bash\n"
@@ -828,6 +831,7 @@ class LabManager:
             env.append(env_var)
 
         # Specification for the user's container.
+        homedir = self._get_homedir(user)
         container = V1Container(
             name="notebook",
             args=["/opt/lsst/software/jupyterlab/runlab.sh"],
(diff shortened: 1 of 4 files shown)


src/jupyterlabcontroller/services/lab.py
```diff
@@ -828,6 +831,7 @@ def build_pod_spec(
             env.append(env_var)
 
         # Specification for the user's container.
+        homedir = self._get_homedir(user)
```
rra, line 834: No need for the temporary variable.

src/jupyterlabcontroller/services/lab.py
```diff
@@ -858,7 +862,7 @@ def build_pod_spec(
                 run_as_group=user.gid,
             ),
             volume_mounts=mounts,
-            working_dir=f"/home/{user.username}",
+            working_dir=homedir,
```
rra, line 865:
```suggestion
            working_dir=self._get_homedir(user),
```

src/jupyterlabcontroller/services/lab.py
```diff
@@ -874,7 +878,8 @@ def build_pod_spec(
             restart_policy="OnFailure",
             security_context=V1PodSecurityContext(
                 run_as_non_root=True,
-                fs_group=user.gid,
```
rra, line 877: `fs_group` has to be set or the user won't have write access to `/tmp`, I think. Although if you've tested and this is not the case, that's good news.

src/jupyterlabcontroller/services/lab.py
```diff
@@ -874,7 +878,8 @@ def build_pod_spec(
...
+                run_as_user=user.uid,
+                run_as_group=user.gid,
```
rra, line 882: These were previously set in the container rather than the pod level. I'm okay with setting them here instead and overriding them at the container level for init containers, but this needs verification that they can be overridden at the container level when they're set at the pod level (in other words, that this doesn't break privileged init containers).

Also, I'm not sure this does anything. I think the assumption here is that the creation of the pod working directory would be done as the user configured in the pod security context, but is that documented anywhere? It would make more sense to me from a security standpoint if the working directory were created by the Kubernetes control plane, and thus is always created as root regardless of the pod security settings.

rra (commented):
This doesn't seem like an emergency, and I don't think we should merge changes without tests where it seems reasonably straightforward to write the test. We already have a configuration for the other user home directory scheme in `tests/configs/homedir-schema`. So testing this is mostly a matter of cloning the `test_lab_objects` test in `tests/handlers/labs_test.py` but use an application configured with that directory instead of the default.

We haven't set up a good way to run the application with alternate configurations yet, so that's the hard part, and probably what you're referring to. This shouldn't require new helper code, though; it should just require a way to reconfigure the running application. I think calling `context_dependency.aclose()` followed by `config_dependency.set_path()` and then `context_dependency.initialize()` on the new configuration object would do it?

src/jupyterlabcontroller/services/lab.py
athornton:
```
[athor@athor-nb ~]$ cd /tmp
[athor@athor-nb tmp]$ cat > bar
foo
[athor@athor-nb tmp]$ ls
bar
[athor@athor-nb tmp]$ cat bar
foo
[athor@athor-nb tmp]$ 
```

src/jupyterlabcontroller/services/lab.py
athornton: Container overrides Pod:

```
To specify security settings for a Container, include the securityContext field in the Container manifest. The securityContext field is a [SecurityContext](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#securitycontext-v1-core) object. Security settings that you specify for a Container apply only to the individual Container, and they override settings made at the Pod level when there is overlap. Container settings do not affect the Pod's Volumes.
```

cf https://kubernetes.io/docs/tasks/configure-pod-container/security-context/

But....I think you're right that it doesn't do anything for us here.

athornton: Dropping fs_group still seems worthwhile, though.

rra: Yeah, no objections to putting this in at the pod level. I'm not sure if it does anything, but it can't hurt.

src/jupyterlabcontroller/services/lab.py
rra: Oh, great. Then yes, by all means let's drop that. I had looked at doing that before because we don't really want `fs_group` mounting semantics, but had convinced myself it broke things when removed.

athornton:
I think that's it except for the test case, which I'll figure out tomorrow.

athornton:
My bad, didn't mean to close.

athornton:
OK.  There's a test now.  PodSpec isn't JSON-serializable, so rather than beat my head on that, I just directly looked at the value of the field (.containers[0].workingDir) that I care about in the test case.  Hope that's OK.

Likewise, I didn't go through all the work to start a prepuller and reconcile images and turn a textual size description into LabResources, I just instantiated an RSPImage and LabResources directly to use in the Pod construction.

rra:
I wasn't sure if I was missing something when I thought the test was straightforward, so I checked out your branch and tried to write it and the following seems to work. Could we test this way instead of this much more complicated test that pokes around at the internals in a way that I think will be less maintainable, since all those details are subject to later refactoring?

```python
@pytest.mark.asyncio
async def test_pod_working_dir(
    client: AsyncClient,
    mock_kubernetes: MockKubernetesApi,
    obj_factory: TestObjectFactory,
) -> None:
    """Check that the pod working directory uses the right home directory.

    Earlier versions had a bug where the working directory for the spawned pod
    was always :file:`/home/{username}` even if another home directory rule
    was set.
    """
    token, user = obj_factory.get_user()
    lab = obj_factory.labspecs[0]

    # Reconfigure the app to use a different home directory scheme.
    await context_dependency.aclose()
    config = configure("homedir-schema")
    await context_dependency.initialize(config)

    r = await client.post(
        f"/nublado/spawner/v1/labs/{user.username}/create",
        json={"options": lab.options.dict(), "env": lab.env},
        headers={
            "X-Auth-Request-Token": token,
            "X-Auth-Request-User": user.username,
        },
    )
    assert r.status_code == 201

    namespace = f"{config.lab.namespace_prefix}-{user.username}"
    pod_name = f"{user.username}-nb"
    pod = await mock_kubernetes.read_namespaced_pod(pod_name, namespace)
    for container in pod.spec.containers:
        assert container.working_dir == "/home/r/rachel"
```

athornton:
Sure, I'll do that.  It just felt weird to do the whole end-to-end test just to get at one bit of the data.

athornton:
I moved it over to handlers in its own file, since it's certainly not a service test anymore.

rra (approved):

PR #150, opened 2023-08-21
Force volume object names to lowercase (needed by T&S)

athornton:


athornton:
Working at base, so I have high confidence it will work at TTS/Summit.

rra (approved):

PR #158, opened 2023-09-08
DM-40691: Add a new generic Kubernetes object watcher

rra:
Add a generic Kubernetes watcher class with proper typing for the objects that it returns, built on the assumption that the Kubernetes library will properly decode Kubernetes objects in the watch events. For now, use it only to watch for pod status; the other uses of watches in the Kubernetes storage layer will be replaced in upcoming changes.

After some experimentation, this uses a different approach than the callback approach currently used in the storage layer for watches. Instead, the watch exposes an async interator with structured events with appropriate typing, allowing any analysis to be done in the body of the async for loop.

### DIFF
--- a/src/jupyterlabcontroller/storage/kubernetes/watcher.py
+++ b/src/jupyterlabcontroller/storage/kubernetes/watcher.py
@@ -1,0 +1,218 @@
+"""Watch a Kubernetes namespace or cluster for events."""
+
+from __future__ import annotations
+
+from collections.abc import AsyncIterator, Awaitable, Callable
+from dataclasses import dataclass
+from datetime import timedelta
+from typing import Any, Generic, Self, TypeVar
+
+from kubernetes_asyncio.client import ApiException
+from kubernetes_asyncio.watch import Watch
+from structlog.stdlib import BoundLogger
+
+from ...exceptions import KubernetesError
+from ...models.domain.kubernetes import WatchEventType
+
+#: Type of Kubernetes object being watched (`dict` for custom objects).
+T = TypeVar("T")
+
+__all__ = [
+    "KubernetesWatcher",
+    "T",
+    "WatchEvent",
+]
+
+
+@dataclass
+class WatchEvent(Generic[T]):
+    """Parsed event from a Kubernetes watch.
+
+    This model is intended only for use within the Kubernetes storage layer.
+    It is interpreted by the callers of the generic watch class inside the
+    per-object Kubernetes storage classes.
+    """
+
+    action: WatchEventType
+    """Action the event represents."""
+
+    object: T
+    """Affected Kubernetes object."""
+
+    @classmethod
+    def from_event(cls, event: dict[str, Any], object_type: type[T]) -> Self:
+        """Create a `KubernetesWatchEvent` from a watch event.
+
+        Parameters
+        ----------
+        event
+            Event as returned by the Kubernetes watch API.
+        object_type
+            Expected type of the object.
+
+        Raises
+        ------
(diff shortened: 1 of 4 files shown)


src/jupyterlabcontroller/storage/kubernetes/watcher.py
```diff
@@ -0,0 +1,221 @@
...
+
+__all__ = [
+    "KubernetesWatcher",
+    "T",
+    "WatchEvent",
+]
+
+
+@dataclass
+class WatchEvent(Generic[T]):
```
athornton, line 28: Oh no, I'm going to have to lose my fear of Generics and TypeVars, aren't I?

src/jupyterlabcontroller/storage/kubernetes/watcher.py
```diff
@@ -0,0 +1,221 @@
...
+        obj = event["object"]
+        if not isinstance(obj, object_type):
+            real_type = type(obj).__name__
+            expected_type = object_type.__name__
+            msg = f"Watch object was of type {real_type}, not {expected_type}"
+            raise ValueError(msg)
+        return cls(action=action, object=obj)
+
+
+class KubernetesWatcher(Generic[T]):
```
athornton, line 70: So `T` here is built-in to the type system (or maybe `Generic[T]`) and doesn't need any further hinting?

src/jupyterlabcontroller/storage/kubernetes/watcher.py
```diff
@@ -0,0 +1,221 @@
...
+
+    This model is intended only for use within the Kubernetes storage layer.
+    It is interpreted by the callers of the generic watch class inside the
+    per-object Kubernetes storage classes.
+    """
+
+    action: WatchEventType
+    """Action the event represents."""
+
+    object: T
```
athornton, line 39: So...T is just _some_ concrete type, from line 18?

src/jupyterlabcontroller/storage/kubernetes/watcher.py
```diff
@@ -0,0 +1,221 @@
...
+            expected_type = object_type.__name__
+            msg = f"Watch object was of type {real_type}, not {expected_type}"
+            raise ValueError(msg)
+        return cls(action=action, object=obj)
+
+
+class KubernetesWatcher(Generic[T]):
+    """Watch Kubernetes for events.
+
+    This wrapper around the watch API of the Kubernetes client implements
```
athornton, line 73: Eesh.  It feels like the low-effort upstreamable fix here would be to add an optional parameter to kubernetes_asyncio: if it's None, use the existing scary behavior of parsing the docstring, but if it's specified, trust the caller and use that.

No, I don't know if this would actually be low-effort.

athornton (approved):
This seems good, although I'm frightened of Generics and TypeVars, but that's on me.

src/jupyterlabcontroller/storage/kubernetes/watcher.py
rra: That is indeed how kubernetes_asyncio works. If you pass a type to the `Watch` constructor, then it uses that type; otherwise, it does the docstring parsing. That's one of the things that ths wrapper does: pass the explicit type into the constructor.

src/jupyterlabcontroller/storage/kubernetes/watcher.py
rra: Yes, exactly. You can define a bound on the type when you define the `TypeVar`, but in this case it can be any type.

PR #159, opened 2023-09-08
DM-40691: Add new generic Kubernetes storage layer

rra:
Add a couple of classes that provide generic Kubernetes storage access, one to types where we only do create and read and another for types where we also do list and delete. Support deleting the old resource on conflicts for the latter type. Instantiate classes for ConfigMap and Service management from those generic classes and use them in the Kubernetes storage layer.

### DIFF
--- a/src/jupyterlabcontroller/constants.py
+++ b/src/jupyterlabcontroller/constants.py
@@ -9,6 +9,7 @@ __all__ = [
     "DROPDOWN_SENTINEL_VALUE",
     "GROUPNAME_REGEX",
     "IMAGE_REFRESH_INTERVAL",
+    "KUBERNETES_DELETE_TIMEOUT",
     "KUBERNETES_REQUEST_TIMEOUT",
     "LAB_STATE_REFRESH_INTERVAL",
     "LIMIT_TO_REQUEST_RATIO",
@@ -28,6 +29,15 @@ DROPDOWN_SENTINEL_VALUE = "use_image_from_dropdown"
 
 IMAGE_REFRESH_INTERVAL = timedelta(minutes=5)
 """How frequently to refresh the list of remote and cached images."""
+
+KUBERNETES_DELETE_TIMEOUT = timedelta(seconds=60)
+"""How long to wait for deletion of an object to finish.
+
+In some cases, if a Kubernetes object the controller is trying to create
+already exists, it deletes that object and then retries the creation. This
+controls how long it waits for the object to go away after deletion before it
+gives up.
+"""
 
 LAB_STATE_REFRESH_INTERVAL = timedelta(minutes=60)
 """How frequently to refresh user lab state from Kubernetes.
--- a/src/jupyterlabcontroller/storage/k8s.py
+++ b/src/jupyterlabcontroller/storage/k8s.py
@@ -55,6 +55,8 @@ from ..models.domain.kubernetes import (
 )
 from ..models.v1.lab import ResourceQuantity
 from ..util import deslashify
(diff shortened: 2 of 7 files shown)


src/jupyterlabcontroller/constants.py
```diff
@@ -29,6 +30,15 @@
 IMAGE_REFRESH_INTERVAL = timedelta(minutes=5)
 """How frequently to refresh the list of remote and cached images."""
 
+KUBERNETES_DELETE_TIMEOUT = timedelta(seconds=60)
```
athornton, line 33: Is this long enough for namespaces?  I guess if there's a running pod and you haven't said otherwise, it's 30s, so probably so.  We can always change it if we need to, I guess.

src/jupyterlabcontroller/storage/k8s.py
```diff
@@ -110,6 +112,9 @@ def __init__(
         self._method_map = KubernetesKindMethodMapper()
         self._fill_method_map()
 
+        self._config_map = ConfigMapStorage(self.k8s_api, logger)
```
athornton, line 115: All of the k8s wrapping is basically to glue our logging and exceptions to the underlying k8s objects?

src/jupyterlabcontroller/storage/kubernetes/creator.py
```diff
@@ -0,0 +1,157 @@
...
+    api_client
+        Kubernetes API client.
+    logger
+        Logger to use.
+    """
+
+    def __init__(self, api_client: ApiClient, logger: BoundLogger) -> None:
+        api = client.CoreV1Api(api_client)
+        super().__init__(
+            create_method=api.create_namespaced_config_map,
```
athornton, line 152: OK, there it is.

src/jupyterlabcontroller/storage/kubernetes/creator.py
```diff
@@ -0,0 +1,157 @@
...
+T = TypeVar("T", bound=KubernetesModel)
+
+__all__ = [
+    "ConfigMapStorage",
+    "KubernetesObjectCreator",
+    "T",
+]
+
+
+class KubernetesObjectCreator(Generic[T]):
```
athornton, line 36: We also deal with Volumes, which are not namespaced, so I feel like you may want to jam "Namespaced" into the class name, or have "NonNamespaced" variants of these for Volumes and any other non-namespaced objects we deal with.

athornton: This is essentially the same problem we dealt with way back when with the Kubespawner Reflectors.

src/jupyterlabcontroller/storage/k8s.py
athornton: Yes, having read to the end, it is.

athornton (approved):
I'm not convinced yet that adding this abstraction layer is a good trade-off, but I suspect that when I see how tests and mocks get a lot simpler I will be convinced.

Right now this seems like it adds a fairly confusing layer of abstraction and turning API calls into object methods that makes the code more difficult for me to understand, but I trust that your reasoning is sound and in the next few PRs I will have the lightbulb moment.

src/jupyterlabcontroller/constants.py
rra: It's not, but namespaces will get their own timeout handling. I'm not entirely happy with this being a constant, though, and timeouts were a bit messy. I have it on my list to go back and take another pass at timeout handling later.

PR #162, opened 2023-09-12
DM-40691: Use generic Kubernetes storage for more kinds

rra:
Use the generic storage layer to provide storage for Secret, NetworkPolicy, PersistentVolumeClaim, and ResourceQuota objects.


athornton (approved):
Making sense.  At this point it is still feeling like it's not necessarily a net reduction in complexity, but it's not growing either, and maybe being centralized?

rra:
Yeah, the main merit from my perspective is that I can fix bugs only once. Any changes to how we wait for deletion, or how we handle timeouts, or what information we want in exceptions or log messages can be done in a small handful of places instead of every place we manage a Kubernetes resource, and the higher level calls all use a uniform syntax.

PR #163, opened 2023-09-12
DM-40691: Add a new-style NamespaceStorage class

rra:
Replace the direct calls to the Kubernetes namespace API with a new NamespaceStorage class in the new style. This class is entirely separate from those based on KubernetesObjectDeleter even though there is some duplication of code, since the namespace API is fundamentally different (only a name, no namespace).

Preserve the more complex timeout handling for namespace deletion since it is most likely to have timeout issues. This may be worth implementing for all deleter objects, but that's a problem for a future day.


rra:
This and custom objects (coming next) do introduce some code duplication, unfortunately, but I looked at trying to make the layer generic enough to handle them and apart from watches it just wasn't worth it. There are differences in logging, exceptions, and the API calls, so it's more straightforward to just have separate code.

athornton (approved):
LGTM

PR #164, opened 2023-09-12
DM-40691: Add new JobStorage and IngressStorage classes

rra:
Replace the Kubernetes job manipulation classes with a JobStorage class that inherits from KubernetesObjectDeleter. Add propagation_policy support to the standard delete method to support its usage with jobs.

Replace the Kubernetes ingress calls with a new-style IngressStorage class. This inherits the standard functionality from the generic KubernetesObjectDeleter class, but adds a method to check whether an ingress has an IP address assigned and a method to wait for an ingress to appear and get an IP address.

This uncovered a bug that meant that the code waiting for the ingress and job to be deleted was doing nothing due to a missing f in an f-string. Fixing that in turn required some changes to the test suite, since it was often not deleting the ingress until after the controller was waiting for the ingress to go away.

### DIFF
--- a/src/jupyterlabcontroller/storage/k8s.py
+++ b/src/jupyterlabcontroller/storage/k8s.py
@@ -49,6 +49,7 @@ from ..models.domain.kubernetes import (
     KubernetesPodEvent,
     KubernetesPodPhase,
     KubernetesPodWatchInfo,
+    PropagationPolicy,
     WatchEventType,
     get_watch_args,
 )
@@ -61,7 +62,8 @@ from .kubernetes.creator import (
     ResourceQuotaStorage,
     SecretStorage,
 )
-from .kubernetes.deleter import ServiceStorage
+from .kubernetes.deleter import JobStorage, ServiceStorage
+from .kubernetes.ingress import IngressStorage
 from .kubernetes.namespace import NamespaceStorage
 from .kubernetes.watcher import KubernetesWatcher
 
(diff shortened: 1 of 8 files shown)


src/jupyterlabcontroller/storage/k8s.py
```diff
@@ -1421,11 +1367,11 @@ async def wait_for_fileserver_object_deletion(
         happens basically immediately after its corresponding Ingress
         is deleted.
         """
-        name = "{username}-fs"
```
athornton, line 1424: Oops.

src/jupyterlabcontroller/storage/k8s.py
```diff
@@ -1453,68 +1399,7 @@ async def get_fileserver_pod_for_user(
...
-        )
-        if ingress is None:
-            # Well, OK, but we know what it's going to look like, so we
-            # can do the watch anyway...
-            expected_metadata = V1ObjectMeta(name=name, namespace=namespace)
-            watch_args = get_watch_args(expected_metadata)
-        else:
-            watch_args = get_watch_args(ingress.metadata)
-        # Do we already have a working ingress?
-        if (
```
athornton, line 1470: I definitely like the new version better.

src/jupyterlabcontroller/storage/k8s.py
```diff
@@ -1453,68 +1399,7 @@ async def get_fileserver_pod_for_user(
...
-            ingress is not None
-            and ingress.status is not None
-            and ingress.status.load_balancer is not None
-            and ingress.status.load_balancer.ingress is not None
-            and len(ingress.status.load_balancer.ingress) > 0
-            and ingress.status.load_balancer.ingress[0].ip != ""
-        ):
-            return
-
-        def transmogrifier(event: dict[str, Any]) -> bool:
```
athornton, line 1480: And good riddance!

athornton (approved):
Now the benefits of the new approach are beginning to become apparent.

PR #166, opened 2023-09-12
DM-40691: Add a new-style CustomStorage class

rra:
Add a new storage object for Kubernetes custom objects, and use it to generate a storage object for GafaelfawrIngress objects. Use that instead of direct calls to the custom object API.

### DIFF
--- a/src/jupyterlabcontroller/storage/kubernetes/custom.py
+++ b/src/jupyterlabcontroller/storage/kubernetes/custom.py
@@ -1,0 +1,303 @@
+"""Storage layer for Kubernetes custom objects objects."""
+
+from __future__ import annotations
+
+from typing import Any
+
+from kubernetes_asyncio import client
+from kubernetes_asyncio.client import ApiClient, ApiException
+from structlog.stdlib import BoundLogger
+
+from ...constants import KUBERNETES_DELETE_TIMEOUT
+from ...exceptions import KubernetesError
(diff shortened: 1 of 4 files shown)


src/jupyterlabcontroller/storage/kubernetes/custom.py
```diff
@@ -0,0 +1,303 @@
...
+                return None
+            raise KubernetesError.from_exception(
+                "Error reading object",
+                e,
+                kind=self._kind,
+                namespace=namespace,
+                name=name,
+            ) from e
+
+    async def wait_for_deletion(self, name: str, namespace: str) -> None:
```
athornton, line 202: I don't know why wait_for_deletion bugs me more than the rest of the duplicated stuff.  Maybe it's that watch logic that seems to be repeated almost verbatim.

athornton (approved):
Seems fine.  I feel like there should be some way to factor the deletion watch out into a common something, but if there were you probably would have done it.  I realize every object type needs its own list method.

src/jupyterlabcontroller/storage/kubernetes/custom.py
rra: Yeah, I think I might be able to move some of the logic into the watcher class, although not the construction of the watcher object itself. I'll do that in an upcoming PR.

PR #167, opened 2023-09-13
DM-40691: Add a new pod storage layer

rra:
Add a new PodStorage class based on KubernetesObjectDeleter. Move a generic function for waiting for pod phases to it, and call that function instead of doing the watch directly in K8sStorageClient. Move the async iterator for pod events into that class as well, and change it to use the new KubernetesWatcher class.

Pods had several features that other types didn't need, so add them to the generic layer: proper timeout tracking on watches in case the server times us out before our timeout was reached, grace period on object deletion, and listing with a label selector.

Use a custom logger for watches from the new storage layer so that any log messages produced by the watch will include the object name and namespace when appropriate. Also fix the KubernetesWatcher object to not destructively modify its arguments so that it's reusable, not that we use that at the moment.

Clean up all of the old generic storage infrastructure from K8sStorageClient, since it's no longer needed.

### DIFF
--- a/src/jupyterlabcontroller/factory.py
+++ b/src/jupyterlabcontroller/factory.py
@@ -15,7 +15,6 @@ from safir.slack.webhook import SlackWebhookClient
 from structlog.stdlib import BoundLogger
 
 from .config import Config
-from .constants import KUBERNETES_REQUEST_TIMEOUT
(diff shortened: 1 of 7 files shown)


athornton (approved):
Nice.  Almost 300 lines lighter.

PR #168, opened 2023-09-13
DM-40691: Use label selector for listing jobs

rra:
When reconstructing file server state, we want to retrieve all jobs with a given label. Use a label selector so that Kubernetes does the work for us.


athornton (approved):
Good catch.

PR #175, opened 2023-09-19
DM-40813: Refactor prepuller storage layer

rra:
Model the new separation between builder objects, services, and the Kubernetes storage layer with the prepuller, which is the simplest. Add a new PrepullerBuilder that constructs the pod that the prepuller uses, and have it use that and the PodStorage storage layer directly, so it no longer needs K8sStorageClient.

Introduce a new MetadataStorage layer, which is resposible for getting metadata about the running pod from the Kubernetes downward API, and use it for namespace and object reference information for the prepuller pods.

Stop using the namespace prefix for labs as the namespace for prepuller pods and always use the namespace from the downward API. Make the downward API information mandatory; there's really no reason not to have it.

Stop adding Argo CD annotations to the prepuller pods. Since they are children of the controller pod, they don't need any.

### DIFF
--- a/src/jupyterlabcontroller/storage/metadata.py
+++ b/src/jupyterlabcontroller/storage/metadata.py
@@ -1,0 +1,55 @@
+"""Storage layer for pod metadata."""
+
+from __future__ import annotations
+
+from functools import cached_property
+from pathlib import Path
+
+from kubernetes_asyncio.client import V1OwnerReference
+
+__all__ = ["MetadataStorage"]
+
+
+class MetadataStorage:
+    """Storage layer for pod metadata for the running pod.
+
+    Some attributes of the Kubernetes objects created by the lab controller
+    depend on metadata about where the lab controller is running. This storage
+    layer retrieves that information via the Kubernetes `downward API`_.
+
+    Parameters
+    ----------
+    metadata_path
+        Path at which the downward API data is mounted.
+    """
+
+    def __init__(self, metadata_path: Path) -> None:
+        self._path = metadata_path
+
+    @cached_property
+    def namespace(self) -> str:
+        """The namespace in which the lab controller is running.
+
+        Some resources, such as prepuller pods, are spawned in the same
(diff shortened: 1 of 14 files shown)


src/jupyterlabcontroller/storage/metadata.py
```diff
@@ -0,0 +1,55 @@
...
+        return (self._path / "namespace").read_text().strip()
+
+    @cached_property
+    def owner_reference(self) -> V1OwnerReference:
+        """An owner reference pointing to the lab controller.
+
+        We want some objects, such as prepuller pods, to show as owned by the
+        lab controller. This enables clearer display in services such as Argo
+        CD and also hopefully tells Kubernetes to delete the pods when the lab
+        controller is deleted, avoiding later conflicts.
```
athornton, line 45: Oh, nice.  Yeah, that was kind of a pain to have to manage manually.

tests/configs/standard/input/metadata/namespace
```diff
@@ -0,0 +1 @@
+nublado
```
athornton, line 1: Don't really understand the motivation here, but no objection.

athornton (approved):
All looks good.  It took me a bit to understand what the testcases were doing but once I saw how it fit with the DownwardAPI mount it made sense.

rra:
Yeah, the idea is that we were previously getting the controller's namespace from the magic file in its service account, but (a) we were using downward API already for everything else, and (b) this assumes that there will always be a service account available and the service account metadata will include the namespace, which is likely but not guaranteed. Since we were doing downward API anyway, I just added the namespace to the downward API specification for the controller pod and then did everything the same way.

The test suite changes are because previously we had a fallback to put the prepuller pods in userlabs (which didn't really make sense) if no service account was available, and we had fallbacks for omitting the owner reference if downward API data wasn't available. But there's no reason for it to not be available, so I just made it use downward API unconditionally, which required adding it to a few more tests.

PR #178, opened 2023-09-20
DM-40813: Move node storage into its own class

rra:
Continuing the disassembly of K8sStorageClient into smaller classes, move get_image_data to get image information from that class into a new NodeStorage class, which is now the only Kubernetes storage required by the ImageService.


athornton (approved):
LGTM.

PR #179, opened 2023-09-20
DM-40813: Move all lab object creation into a builder object

rra:
Move the builder object for the prepuller into a separate module and move the lab builder into services.builder.lab. Rework how lab creation is done to move all Kubernetes object construction into the builder, which returns a dataclass containing all the objects. Add a new lab-specific storage layer that's responsible for making the calls to create those objects, plus some ancillary methods for secret management.

Move the merging of secret data into the lab service, where it seems to fit better, although there's no ideal location for it.

Make the Argo CD application to which lab objects are attached configurable, rather than hard-coding it. Move a bunch of paths inside the lab container into constants in preparation for making them configurable.

Simplify the naming of volumes inside the pod specification. These are all scoped to the pod, so there's no reason to add a username-nb prefix to all of them. Also simplify the naming of files in the file config map to use only the file name instead of the full path, since the internal naming of the volumes already assumed that the file name was unique. If we need to support files of the same name in multiple locations in the future, we can cross that bridge then. Since they already weren't working, take the readability win.

Ensure that the nublado.lsst.io/user annotation is set uniformly on all lab objects.

Rename the helper functions for reading JSON test data to explicitly read JSON data so that we can add one that reads plain files. Merge a few home directory test cases and put them with the other lab tests.


src/jupyterlabcontroller/constants.py
```diff
@@ -53,6 +57,12 @@
 gives up.
 """
 
+LAB_COMMAND = "/opt/lsst/software/jupyterlab/runlab.sh"
```
athornton, line 60: Interesting.  I can no longer remember why I rely instead on setting things like NONINTERACTIVE inside the lab.

src/jupyterlabcontroller/constants.py
```diff
@@ -72,16 +82,34 @@
 so copy its behavior.
 """
 
+METADATA_PATH = Path("/etc/podinfo")
+"""Default path to injected pod metadata."""
+
+MOUNT_PATH_DOWNWARD_API = "/opt/lsst/software/jupyterlab/runtime"
```
athornton, line 88: I feel like we ought to have the main configurable as some name for /opt/lsst/software and then tack things on to that.  (and maybe make it more generic at the same time)

src/jupyterlabcontroller/exceptions.py
```diff
@@ -519,9 +519,26 @@ def to_slack(self) -> SlackMessage:
...
 
-    def __init__(self, message: str, namespace: str, name: str) -> None:
+    Parameters
+    ----------
+    name
+        Name of secret.
+    namespace
+        Namespace of secret.
+    key
+        Key within that secret that was found if the secret was present but
```
athornton, line 531: I find the wording of the comment very confusing, although what it's doing is quite clear from the code.  Maybe when the coffee kicks in I will have a constructive solution.

src/jupyterlabcontroller/services/builder/lab.py
```diff
@@ -0,0 +1,821 @@
...
+        volumes: list[LabVolume],
+        prefix: str = "",
+    ) -> list[LabVolumeContainer]:
+        """Construct volumes and mounts for volumes in the lab configuration.
+
+        Parameters
+        ----------
+        username
+            Name of user this lab is for, used to name the PVC objects.
+        volumes
```
athornton, line 173: Good call on specifically calling this out in a comment.

src/jupyterlabcontroller/services/builder/lab.py
```diff
@@ -0,0 +1,821 @@
...
+
+        Returns
+        -------
+        list of LabVolumeContainer
+            List of volumes and mounts.
+        """
+        results = []
+        pvc_count = 1
+        for storage in volumes:
+            is_readonly = storage.mode == FileMode.RO
```
athornton, line 189: Obviously we don't need them, but a set of parens here would help readability for me.

src/jupyterlabcontroller/services/builder/lab.py
```diff
@@ -0,0 +1,821 @@
...
+                            server=source.server,
+                        ),
+                    )
+                case PVCVolumeSource():
+                    claim = V1PersistentVolumeClaimVolumeSource(
+                        claim_name=f"{username}-nb-pvc-{pvc_count}",
+                        read_only=is_readonly,
+                    )
+                    volume = V1Volume(name=name, persistent_volume_claim=claim)
+                    pvc_count += 1
```
athornton, line 212: It feels so weird without a default case!

src/jupyterlabcontroller/services/builder/lab.py
```diff
@@ -0,0 +1,821 @@
...
+        dict of str
+            Equivalent environment sent by JupyterHub.
+
+        Notes
+        -----
+        The list of environment variables that are always added internally by
+        the lab controller must be kept in sync with the code that creates the
+        config map.
+        """
+        unwanted = {
```
athornton, line 264: I always forget set literals are a thing.

src/jupyterlabcontroller/services/builder/lab.py
```diff
@@ -0,0 +1,821 @@
...
+        }
+        return {k: v for k, v in env.items() if k not in unwanted}
+
+    def _build_home_directory(self, username: str) -> str:
+        """Construct the home directory path for a user."""
+        match self._config.homedir_schema:
+            case UserHomeDirectorySchema.USERNAME:
+                return f"/home/{username}"
+            case UserHomeDirectorySchema.INITIAL_THEN_USERNAME:
+                return f"/home/{username[0]}/{username}"
```
athornton, line 286: Should we just go ahead and parameterize /home while we're in here, since we're about to need it?

src/jupyterlabcontroller/services/builder/lab.py
```diff
@@ -0,0 +1,821 @@
...
+            ),
+            # Node on which the pod is running.
+            V1EnvVar(
+                name="KUBERNETES_NODE_NAME",
+                value_from=V1EnvVarSource(
+                    field_ref=V1ObjectFieldSelector(field_path="spec.nodeName")
+                ),
+            ),
+            # Deprecated version of KUBERNETES_NODE_NAME, used by lsst.rsp
+            # 0.3.4 and earlier.
```
athornton, line 784: We really need to prioritize a reaper for old images, so that we know when anything that uses 0.3.4 has gone away, so we can clean up this (and other) tech debt items.

src/jupyterlabcontroller/services/state.py
```diff
@@ -150,7 +150,7 @@ async def get_lab_state(self, username: str) -> UserLabState:
             phase = await self._kubernetes.get_pod_phase(pod, namespace)
         except KubernetesError as e:
             self._logger.exception(
-                "Cannot get pod status",
+                "Cannot get pod phase",
```
athornton, line 153: Good catch.

src/jupyterlabcontroller/storage/kubernetes/lab.py
```diff
@@ -0,0 +1,141 @@
...
+    Parameters
+    ----------
+    api_client
+        Kubernetes API client.
+    logger
+        Logger to use.
+
+    Notes
+    -----
+    This class isn't strictly necessary; instead, the lab service could
```
athornton, line 37: I was just wondering, "why are we..." so +1 for the comment.

tests/configs/homedir-schema/input/config.yaml
```diff
@@ -8,59 +8,10 @@ lab:
...
-        halt:x:7:0:halt:/sbin:/sbin/halt
-        mail:x:8:12:mail:/var/spool/mail:/sbin/nologin
-        operator:x:11:0:operator:/root:/sbin/nologin
-        games:x:12:100:games:/usr/games:/sbin/nologin
-        ftp:x:14:50:FTP User:/var/ftp:/sbin/nologin
-        tss:x:59:59:Account used by the trousers package to sandbox the tcsd daemon:/dev/null:/sbin/nologin
-        dbus:x:81:81:System message bus:/:/sbin/nologin
-        nobody:x:99:99:Nobody:/:/sbin/nologin
-        systemd-network:x:192:192:systemd Network Management:/:/sbin/nologin
-        lsst_lcl:x:1000:1000::/home/lsst_lcl:/bin/bash
```
athornton, line 26: Not sure why we wouldn't just leave the default passwd/group files in the tests, but OK.

tests/configs/homedir-schema/output/passwd
```diff
@@ -0,0 +1,2 @@
+root:x:0:0:root:/root:/bin/bash
```
athornton, line 1: OK, I guess this is why, and it's simpler, yes.

tests/services/state_test.py
```diff
@@ -58,38 +58,33 @@ async def create_lab(
...
-    await lab_manager.create_secrets(user)
-    await lab_manager.create_nss(user)
-    await lab_manager.create_file_configmap(user)
-    await lab_manager.create_env(user, lab, image)
-    await lab_manager.create_network_policy(user)
-    await lab_manager.create_quota(user)
-    await lab_manager.create_lab_service(user)
-    await lab_manager.create_user_pod(user, resources, image)
+    #
+    # Create this lab with an empty set of seret data, since it shouldn't
```
athornton, line 82: "secret"

athornton (approved):
I think I like the builder concept, although it still feels like we've got irreducible complexity and pushing it one abstraction layer deeper may not be all that helpful--but I won't know until I haven't looked at the controller for six months to a year and then go back to it.  Your instincts are better than mine, so it probably *is* helpful.

This has done a really good job exposing places we've hardcoded particular paths (well, usually paths) that will need to be configurable.  My vote would be to go ahead and do that while we're in here, but I certainly can see the merit in putting it off too.

I like the renaming the *data functions to *json in the test suite, which makes their function clearer.

I do kind of wonder if /opt/lsst/software/jupyterlab and /etc/podinfo and all the rest of our injected not-actual-fileserver-mounts should be consolidated to a single path (/var/run/nublado by default maybe?) and runlab.sh tweaked to accomodate that.  I don't want the default container to be the mess of spaghetti mounts that, for instance T&S base is.

src/jupyterlabcontroller/constants.py
rra: Yes, that would make sense. I'll add that to my list of things to do when we make it configurable. I'm hoping we can also get rid of a few of the things we're mounting. Both the environment variables and the downward API in the user lab are of questionable utility.

src/jupyterlabcontroller/services/builder/lab.py
rra: I tried, but black just removes them again.

src/jupyterlabcontroller/services/builder/lab.py
rra: Type checking is great for that!

src/jupyterlabcontroller/services/builder/lab.py
rra: That's a logically separate change, so I'd rather do it as a separate PR.

PR #181, opened 2023-10-02
DM-40813: Automatically reconfigure running test apps

rra:
If a test version of the controller is running and a test case calls configure, automatically reset the process context to use the new configuration.


athornton (approved):
This makes things conceptually much clearer in the test suite.

PR #182, opened 2023-10-04
DM-40813: Move fileserver object creation into a builder

rra:
Follow the new builder pattern for fileserver object creation and move the Kubernetes object construction into FileserverBuilder. Move construction of Kubernetes classes for volumes and volume mounts into another small class that can be used by both the LabBuilder and the FileserverBuilder. Move most of the storage layer for the fileserver into a new storage class.

The storage object migration is not yet complete because of state management and the calls to reconstruct state from Kubernetes objects. That redesign will be handled in a subsequent redesign of the separation between the manager and state classes to prepare for Redis separation, after which the old Kubernetes storage layer can be retired completely.

This surfaced a problem with the new watcher class. It was rounding down timeouts instead of rounding them up, which caused functions to be called with a zero timeout. Round up explicitly and also error out if the timeout is set but is 0 or less.

### DIFF
--- a/src/jupyterlabcontroller/storage/kubernetes/fileserver.py
+++ b/src/jupyterlabcontroller/storage/kubernetes/fileserver.py
@@ -1,0 +1,179 @@
+"""Kubernetes storage layer for user fileservers."""
+
+from __future__ import annotations
+
+from datetime import timedelta
+
+from kubernetes_asyncio.client import ApiClient, V1Pod
+from safir.datetime import current_datetime
+from structlog.stdlib import BoundLogger
+
+from ...exceptions import DuplicateObjectError, MissingObjectError
+from ...models.domain.fileserver import FileserverObjects
+from ...models.domain.kubernetes import PodPhase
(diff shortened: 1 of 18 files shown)


src/jupyterlabcontroller/storage/kubernetes/fileserver.py
```diff
@@ -0,0 +1,179 @@
...
+        Kubernetes API client.
+    logger
+        Logger to use.
+
+    Notes
+    -----
+    This class isn't strictly necessary; instead, the fileserver service could
+    call the storage layers for individual Kubernetes objects directly. But
+    there are enough different objects in play that adding a thin layer to
+    wrangle the storage objects makes the lab service easier to follow.
```
athornton, line 37: Copy/paste error: "fileserver service"

athornton (approved):
All I could find wrong was one comment.  This looks great.  I will have some questions about whether there's a better workflow for reviewing big refactorings in co-work, though.

PR #185, opened 2023-10-05
DM-41051: Use controller namespace as secret source

rra:
When creating the lab secret, the lab service was still using the namespace prefix for user namespaces as a namespace from which to read secrets, instead of using the namespace of the controller itself. Convert the lab manager to the new MetadataStorage layer so that it is consistent with the rest of the controller, and fix the test suite accordingly.


athornton (approved):
Looks good.

PR #186, opened 2023-10-05
DM-41051: Redo how disabled fileserver is handled

rra:
To make testing easier and allow enabling of the fileserver after the FastAPI application has been started, unconditionally register the fileserver routes but raise exceptions in the fileserver state manager if it's not enabled. Change DisabledError to NotConfiguredError to match the exception for the same purpose in Gafaelfawr and configure it to translate into 404 errors for users.

### DIFF
--- a/src/jupyterlabcontroller/exceptions.py
+++ b/src/jupyterlabcontroller/exceptions.py
@@ -21,7 +21,6 @@ from safir.slack.webhook import SlackIgnoredException
 
 __all__ = [
     "ClientRequestError",
-    "DisabledError",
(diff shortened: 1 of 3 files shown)


athornton (approved):
In practice, would we ever enable/disable the fileserver on the fly?  Making testing easier is definitely a goal I can get behind, though.

rra:
We never would in the running service, but an upcoming change to the test suite does exactly that to make it easier to manage how configurations and fixtures work (basically to work around the problem that you can't easily pass arguments into fixtures).

PR #187, opened 2023-10-06
DM-41051: Refactor the test suite

rra:
Move test data from tests/configs to tests/data following the pattern of other packages. Add a new base directory that contains default input data used by multiple tests, and fall back to that directory for files such as Docker credentials if an individual test does not override that file. Delete the symlinks that were previously used to share test data with the standard test.

Use the data loading functions uniformly rather than reading JSON or text files directly, and add more variations that load data into particular model classes. Replace the remaining pieces of the test objects fixture and test objects JSON with separate files per type of object and those separate functions.

Take advantage of the support for dynamic reconfiguration to remove the separate fileserver fixtures and merge fileserver support into the main helper functions and fixtures. Move the support library for fileserver tests into tests/support with the other support code.


athornton (approved):
I think the basic sentiment, which is to eliminate the TestObjectFactory as a test fixture and replace it with set of functions that just read the input configuration data, is sound.  This is coupled to a more robust way of handling fallback-to-default-data when a test wants to vary just one thing (or a few things) from the default test environment, so we don't have to repeat ourselves all over the place or use a symlink to indicate "this bit of data is unchanged" (the previous behavior).

I did not follow all the file migration necessary to make this happen, but if all the tests pass, I'm happy with it.

PR #188, opened 2023-10-06
DM-41051: Streamline test user handling

rra:
Take advantage of the new `GafaelfawrUser` model to return a single object instead of separate token and user fixtures, and add a method to it to turn a user back into the headers that would be sent by Gafaelfawr as a convenience for tests. Use that everywhere instead of constructing the headers manually except in places where the manual construction adds more test surface.


athornton (approved):
Definite improvement.

PR #191, opened 2023-10-10
tickets/DM-41136: make user name optional

athornton:


rra (approved):
Yup, I think that will work. I thought it was going to be more complicated, but it looks like we already handle the case where that annotation was missing. (Which is interesting, since that should have produced a type error since we may assign name to `None`. Not sure why mypy didn't catch that before this change.)

PR #195, opened 2023-10-11
DM-41051: Add support for homedir prefix and suffix

rra:
Allow the prefix for home directories (normally /home) and the suffix (normally empty, but allowing the home directory for labs to be a subdirectory of the user's normal home directory) to be configured.

I picked this approach over adding template support because I don't like having to document a template language in user documentation and try to explain what variables are available and how to write a template. In this case, I think the three options are reasonably easy to understand and avoid the need for a full-blown templating mini-language.

### DIFF
--- a/src/jupyterlabcontroller/config.py
+++ b/src/jupyterlabcontroller/config.py
@@ -309,7 +309,14 @@ class LabConfig(CamelCaseModel):
     homedir_prefix: str = Field(
         "/home",
         title="Prefix for home directory path",
-        description="Portion of home directory path added before the username",
+        description=(
+            "Portion of home directory path added before the username. This"
+            " is the path *inside* the container, not the path of the volume"
+            " mounted in the container, so it need not reflect the structure"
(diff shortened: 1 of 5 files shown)


src/jupyterlabcontroller/config.py
```diff
@@ -322,6 +332,16 @@ class LabConfig(CamelCaseModel):
         ),
     )
 
+    @field_validator("homedir_prefix")
+    @classmethod
+    def _validate_homedir_prefix(cls, v: str) -> str:
+        return v.rstrip("/")
```
athornton, line 338: Should this not check that the prefix is not "/"?  I don't know who would make user directories the root, but returning the empty string if they do probably isn't the right answer?

athornton (approved):
Looks good.  I was concerned that prefix could be the empty string if someone was mad enough to put homedirs in the root, but the construction of the actual homedir addresses that, so it's fine.

PR #197, opened 2023-10-12
DM-40813: Move lab state reconstruction into builder

rra:
Move the code to reconstruct lab state from objects on disk out of the state manager and into the lab builder so that it lives side by side with the code that creats those objects. In order to maintain the separation between the builder and the storage layer, introduce a couple of new dataclasses, one that holds all the names for user objects and can be passed down to the storage layer, and one that holds just the Kubernetes objects required for state reconstruction.

Tidy up the builder API by removing a few now-unnecessary methods that were too low-level, and clean up more now-unnecessary methods in the old Kubernetes storage layer.


src/jupyterlabcontroller/storage/kubernetes/lab.py
```diff
@@ -93,23 +98,117 @@ async def delete_namespace(self, name: str) -> None:
...
 
         Raises
         ------
         KubernetesError
             Raised if there is some failure in a Kubernetes API call.
         """
         grace_period = LAB_STOP_GRACE_PERIOD
-        await self._pod.delete(name, namespace, grace_period=grace_period)
+        namespace = names.namespace
+        await self._pod.delete(names.pod, namespace, grace_period=grace_period)
```
athornton, line 116: This is just my OCD here, but if you're going to use a temporary variable for either of podname or namespace, please use it for both.

athornton (approved):
That's some nice encapsulation there.  Only one comment, and honestly you can ignore it because it's just one of those things that I am the only person on the planet that would be bothered by it.

PR #198, opened 2023-10-12
DM-40813: Restructure file server state reconciliation

rra:
Use the same pattern for file server state reconciliation as is used for lab state reconciliation: add a method to the underlying storage to retrieve relevant objects, a model that holds those objects, and a builder method to check those objects for correctness. This required access to the function to check whether an Ingress has an IP assigned, so move that into a separate function that can be imported.

Stop checking whether a user has a running file server before creating a new file server. If the user doesn't occur in the map, just create the file server with the recreate flag set. State reconciliation during startup should have found all running file servers, so if the user is not in the map but the objects exist, we're in some error state and should replace the objects anyway.

This retires the last of the old Kubernetes storage layer.


athornton (approved):
LGTM

PR #200, opened 2023-10-25
Refactor lab state management

rra:
The production IDF cluster was seeing a steady stream of KeyError exceptions trying to look up a user's lab in the _labs structure in the LabStateManager. I believe this is due to two general class of race conditions that might cause simultaneous modifications of the state structure. This extensive refactoring attempts to address each of them.

The first is two simultaneous operations for the same user. There was some attempt to prevent this, but it had various race conditions due to time-of-check/time-of-use bugs, made worse by the previous behavior of deleting the user's lab state from the dictionary entirely when the lab is deleted and thus creating a window for two lab creations to race. The protection against simultaneous delete and spawn operations was also lacking.

The second was a general class of races with background reconciliation. The reconciliation thread attempted to exclude labs where an operation was in progress, but there were time-of-check/time-of-use bugs with that exclusion logic as well, also hampered by deleting the lab state when the user's lab was deleted.

I attempted various solutions to this within the existing structure, but the division of responsibilities between the LabManager and LabStateManager classes added a great deal of interface complexity and created code structure problems because of reconciliation living in LabStateManager but lab spawning living in LabManager. I therefore took the approach here of unifying LabManager and LabStateManager to ensure one class was responsible for all overall state management.

Responsibility for ensuring that only one lab operation is in progress at a time, and monitoring operations for exceptions or timeouts and doing error reporting and cleanup, is now handled by a per-user LabMonitor class that uses locking to ensure that a given user cannot have multiple conflicting operations in progress. This allows supporting multiple simultaneous deletes, all of which will block on the same underlying delete operation. Spawn plus delete or multiple spawns will now be rejected with 409 conflict responses.


src/jupyterlabcontroller/exceptions.py
```diff
@@ -581,20 +604,18 @@ def to_slack(self) -> SlackMessage:
             Slack message suitable for posting as an alert.
         """
         message = super().to_slack()
-        obj = f"{self.kind} {self.namespace}" if self.namespace else self.kind
-        message.blocks.append(SlackTextBlock(heading="Object", text=obj))
+        if self.url:
+            text = f"{self.method} {self.url}" if self.method else self.url
```
athornton, line 608: You can save one mention of self.url here at the cost of another vertical line but not really worth it.

src/jupyterlabcontroller/services/lab.py
```diff
@@ -2,64 +2,201 @@
...
+
+class _LabOperation(Enum):
+    """Possible operations on a lab that could be in progress."""
+
+    SPAWN = "spawn"
+    DELETE = "delete"
+
+
+@dataclass
+class _Operation:
```
athornton, line 70: All these coroutines floating around are making me nervous.

src/jupyterlabcontroller/services/lab.py
```diff
@@ -2,64 +2,201 @@
...
+        self._slack = slack_client
         self._logger = logger
 
+        # Background task management.
+        self._scheduler: Scheduler | None = None
+
+        # Mapping of usernames to internal lab state.
+        self._labs: dict[str, _State] = {}
+
+        # Triggered when any background thread monitoring spawn progress
```
athornton, line 183: This is not getting any more reassuring.

src/jupyterlabcontroller/services/lab.py
```diff
@@ -69,175 +206,339 @@ async def create_lab(
...
+            tag = spec.options.image_tag
             image = await self._image_service.image_for_tag_name(tag)
 
         # Check to see if the lab already exists. If so, but it is in a failed
-        # state, delete it first.
-        status = await self._lab_state.get_lab_status(user.username)
-        delete_first = status == LabStatus.FAILED
+        # state, we will delete the previous lab first.
+        state = await self.get_lab_state(user.username)
+        delete_first = bool(state and state.status == LabStatus.FAILED)
```
athornton, line 241: You added a TERMINATED state; do we want to check that here as well?

src/jupyterlabcontroller/services/lab.py
```diff
@@ -69,175 +206,339 @@ async def create_lab(
...
 
-        # Start the spawning process. This also checks for conflicts and
-        # raises an exception if the lab already exists and is not in a failed
-        # state.
+        # If there is any operation already in progress, raise an error.
+        # Similarly, if the lab already exists and is not in a failed state,
```
athornton, line 244: +terminated?

src/jupyterlabcontroller/services/lab.py
```diff
@@ -69,175 +206,339 @@ async def create_lab(
...
+        spawner = self._spawn_lab(
+            user=user,
+            state=state,
+            spec=spec,
+            image=image,
+            events=lab.events,
+            delete_first=delete_first,
         )
+        operation = _Operation(_LabOperation.SPAWN, spawner, state, lab.events)
+        await lab.monitor.monitor(operation, self._spawner_done)
```
athornton, line 278: I think I see: we can't just protect the spawn with a mutex because we want multiple people to spawn at the same time, but a given person has to spawn exactly once.

src/jupyterlabcontroller/services/lab.py
```diff
@@ -69,175 +206,339 @@ async def create_lab(
...
+        #
+        # 1. A deletion is already in progress, in which case we want to wait
+        #    on it with everyone else who is monitoring the deletion.
+        # 2. No operation is in progress, in which case we start the deletion
+        #    and take responsibility for clearing the lab state when it
+        #    finishes.
+        # 3. A different operation (presumably a spawn) is in progress, in
+        #    which case we want to raise an error, since currently we don't
+        #    support aborting a spawn in progress by deleting the lab.
+        if lab.monitor.in_progress == _LabOperation.DELETE:
```
athornton, line 317: Can something change the lab.monitor.in_progress state _while_ the if is being tested?  That is, should you assign lab.monitor.in_progress to a temporary variable just before the if, and use that?  Or is it immaterial?

src/jupyterlabcontroller/services/lab.py
```diff
@@ -69,175 +206,339 @@ async def create_lab(
...
+            Raised if the given user has no lab.
         """
-        callback = partial(self._delete_lab_and_namespace, username, 25, 100)
-        await self._lab_state.stop_lab(username, callback)
+        if username not in self._labs:
+            return None
 
-    async def _delete_lab_and_namespace(
-        self, username: str, start_progress: int, end_progress: int
+        # Grab a copy of the lab state, since we're going to make async calls
```
athornton, line 393: Yeah, this is what I meant about the lab_monitor above.

src/jupyterlabcontroller/services/lab.py
```diff
@@ -69,175 +206,339 @@ async def create_lab(
...
+            lab.state.internal_url = None
+            deleter = self._delete_lab(username, lab.state, lab.events)
+            operation = _Operation(
+                _LabOperation.DELETE, deleter, lab.state, lab.events
             )
-            await self._delete_lab_and_namespace(username, 5, 20)
+            await lab.monitor.monitor(operation)
+            await lab.monitor.wait()
+            if lab.state.status == LabStatus.TERMINATED:
+                lab.state = None
```
athornton, line 331: Ah, OK, so maybe we don't want to check for TERMINATED above since we just remove the state entirely.

src/jupyterlabcontroller/services/lab.py
```diff
@@ -69,175 +206,339 @@ async def create_lab(
...
+
+        return iterator()
+
+    async def get_lab_state(self, username: str) -> UserLabState | None:
+        """Get lab state for a user.
+
+        This method underlies the API called by JupyterHub to track whether
+        the user's lab still exists. We want to update that state after
+        Kubernetes changes faster than our reconciliation cycle, so ask
+        Kubernetes directly for the pod phase each time we're asked for the
```
athornton, line 371: How often is this, potentially?  Do we ever need to worry about it?

src/jupyterlabcontroller/services/lab.py
```diff
@@ -69,175 +206,339 @@ async def create_lab(
...
+        username: str,
+        state: UserLabState,
+        events: AsyncMultiQueue[Event],
+        *,
+        start_progress: int = 25,
+        end_progress: int = 100,
     ) -> None:
         """Delete the user's lab and namespace.
 
-        Currently, this just deletes the namespace and lets that delete the
```
athornton, line 214: Huh.  Was that actually true up until now?

src/jupyterlabcontroller/services/lab.py
```diff
@@ -287,3 +588,591 @@ async def _gather_secret_data(
...
+                    to_monitor.append(username)
+
+        # Second pass: take observed state and create any missing internal
+        # state. This is the normal case after a restart of the lab
+        # controller.
+        for username in set(observed.keys()) - known_users:
+            if username not in self._labs:
+                msg = f"Creating record for user {username} from Kubernetes"
+                self._logger.info(msg)
+                self._labs[username] = _State(
```
athornton, line 767: Does all of this get all gross again when we push state into an external redis, which we would presumably want to access asynchonously?

athornton (approved):
It all _seems_ good--I've asked some questions, but they may be dumb ones, and please don't take them as suggestions that the code is wrong rather than I don't quite understand what's doing.

This is more than a bit tricky, I would say.

src/jupyterlabcontroller/services/lab.py
rra: It was even worse in earlier versions. I tried about four ways of structuring this before I found something that I could wrap my head around.

src/jupyterlabcontroller/services/lab.py
rra: This code doesn't handle labs in a completed state yet. I'm going to tackle that in a separate diff since I think it may require rsp-restspawner changes as well. It just lays a bit of the groundwork for that.

src/jupyterlabcontroller/services/lab.py
rra: Right, so we need per-user mutexes, and it seemed cleaner when I put those in separate classes instead of trying to juggle everything in the LabManager class.

src/jupyterlabcontroller/services/lab.py
rra: The guarantee for async code is that only await can yield control, so as long as you don't await anything, no one else can do things behind your back since only one task is actually running at a time.

src/jupyterlabcontroller/services/lab.py
rra: Right, here we have to grab a copy because we're using await. The other code is set up so that it should never need to call await in the critical section.

src/jupyterlabcontroller/services/lab.py
rra: Every Kubernetes upgrade deletes a bunch of pods, so if we don't do this, all those affected users have to wait for the hour-long reconciliation period before they can spawn again. JupyterHub calls this when the user goes to the spawner page, so this means that the user can recover effectively instantly.

src/jupyterlabcontroller/services/lab.py
rra: No, the comment was just wrong.

src/jupyterlabcontroller/services/lab.py
rra: Yes. We'll have to completely redesign the reconciliation pass when we use Redis for state.

PR #202, opened 2023-10-26
Allow delete to abort an in-progress spawn

rra:
Previously, we returned 409 if someone attempted to delete a spawn that was still in progress. This interferes with cleanup that we want to do in the REST Spawner if JupyterHub was restarted while spawns are in progress. JupyterHub may lose the credentials for those labs and need to stop them so that they can be restarted with correct credentials, but the previous constraint meant that the stop would fail until the spawn completed.

Use the new cancel method on the monitor to stop any spawn in progress and then continue with a delete to clean up any remnant Kubernetes resources. Send a failure event to the event stream of any in-progress operation when the operation is cancelled.

### DIFF
--- a/src/jupyterlabcontroller/services/lab.py
+++ b/src/jupyterlabcontroller/services/lab.py
@@ -311,14 +311,27 @@ class LabManager:
         # 2. No operation is in progress, in which case we start the deletion
         #    and take responsibility for clearing the lab state when it
         #    finishes.
-        # 3. A different operation (presumably a spawn) is in progress, in
-        #    which case we want to raise an error, since currently we don't
-        #    support aborting a spawn in progress by deleting the lab.
+        # 3. A spawn is already in progress, in which case we want to abort
+        #    that spawn and then do a normal delete to remove any remnants.
+        #
+        # Also handle the currently-impossible fourth case of some other
+        # operation in progress. There currently is no other operation, but
+        # do something safe in case one appears in the future.
         if lab.monitor.in_progress == _LabOperation.DELETE:
             await lab.monitor.wait()
-        elif not lab.monitor.in_progress:
-            self._builder.build_object_names(username)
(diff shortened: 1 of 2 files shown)


tests/handlers/labs_test.py
```diff
@@ -560,7 +586,7 @@ async def test_spawn_errors(
     mock_slack: MockSlackWebhook,
 ) -> None:
     lab = read_input_lab_specification_json("base", "lab-specification.json")
-    apis_to_fail = {"read_namespaced_secret"}
+    apis_to_fail = set()
```
athornton, line 589: Is this change unrelated?  If not, I don't understand what reading a namespaced secret has to do with it.

src/jupyterlabcontroller/services/lab.py
```diff
@@ -1015,16 +1028,28 @@ def in_progress(self) -> _LabOperation | None:
     async def cancel(self) -> None:
         """Cancel any existing operation.
 
-        This is called during process shutdown. All operations in progress are
-        cancelled, which may strand random lab state that we will detect and
-        clean up during reconciliation.
+        This is called during process shutdown and to abort a spawn in
```
athornton, line 1031: Wait a minute.  I may have just realized why it's mostly the bots that are exploding.

The JupyterHub UI doesn't let you delete a spawn that's in progress...but we might bypass that with mobu.

athornton (approved):
LGTM, and maybe I understood something new.

tests/handlers/labs_test.py
rra: Sorry, that was me being lazy. Unrelated (but correct) change left over from testing something else.

src/jupyterlabcontroller/services/lab.py
rra: mobu only interacts with JupyterHub, so if JupyterHub doesn't permit it, mobu can't do it.

PR #204, opened 2023-10-26
DM-41350: Properly track terminated pods

rra:
Rather than rolling terminated pods into the error status, track them separately, still allowing them to be automatically deleted when a new spawn is requested. Add some more logging on spawn conflicts to aid in debugging.

This does not yet clean up terminated pods, just enables proper state tracking for them so that they can be cleaned up in a subsequent change.


src/jupyterlabcontroller/models/v1/lab.py
```diff
@@ -60,14 +61,12 @@ class LabStatus(Enum):
     FAILED = "failed"
 
     @classmethod
-    def from_phase(cls, phase: str) -> LabStatus:
+    def from_phase(cls, phase: PodPhase) -> LabStatus:
```
athornton, line 64: I have a bad feeling we're also going to need to consider Reason here.  Lifecycle handling is such a mess.

athornton (approved):
Brace for needing the Reason field too, but, yeah.

PR #205, opened 2023-10-26
DM-41350: Delete not-running labs during reconciliation

rra:
When we encounter a lab in a terminated state during background reconciliation (labs that were OOMKilled, idle-culled, etc.), delete the lab. Do this carefully to avoid conflicts with a user operation that might race with us, since it should get precedence. Similarly clean up labs that are in a failed state.

Convert a pod status of Failed to a lab status of terminated, not failed. Use failed to indicate missing Kubernetes resources or similar errors, and terminated to mean that the pod has stopped for whatever reason.

### DIFF
--- a/src/jupyterlabcontroller/models/v1/lab.py
+++ b/src/jupyterlabcontroller/models/v1/lab.py
@@ -52,7 +52,15 @@ class LabSize(str, Enum):
 
 
 class LabStatus(Enum):
-    """Possible states the user's lab may be in."""
+    """Possible states the user's lab may be in.
+
+    This is not directly equivalent to pod phases. It is instead intended to
+    capture the status of the lab from an infrastructure standpoint,
+    reflecting the current intent of the controller. Most notably, labs that
+    have stopped running for any reason (failure or success) use the
+    terminated status. The failed status is reserved for failed Kubernetes
+    operations or missing or invalid Kubernetes objects.
+    """
 
     PENDING = "pending"
     RUNNING = "running"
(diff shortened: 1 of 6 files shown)


src/jupyterlabcontroller/models/v1/lab.py
```diff
@@ -83,9 +91,9 @@ def from_phase(cls, phase: PodPhase) -> LabStatus:
                 return cls.PENDING
             case PodPhase.RUNNING:
                 return cls.RUNNING
-            case PodPhase.SUCCEEDED:
+            case PodPhase.SUCCEEDED | PodPhase.FAILED:
```
athornton, line 94: That looks good.

tests/services/lab_test.py
```diff
@@ -177,6 +178,39 @@ async def test_reconcile_pending(
...
+    user: GafaelfawrUser,
+    mock_kubernetes: MockKubernetesApi,
+) -> None:
+    namespace = read_input_data("base", "metadata/namespace").strip()
+    for secret in read_input_secrets_json("base", "secrets.json"):
+        await mock_kubernetes.create_namespaced_secret(namespace, secret)
+    mock_kubernetes.initial_pod_phase = PodPhase.SUCCEEDED.value
+    await factory.start_background_services()
+
+    # Create a lab through the controller. It should show up in a terminated.
```
athornton, line 194: remove final period.

athornton (approved):
One typo, otherwise looks good.

src/jupyterlabcontroller/models/v1/lab.py
```diff
@@ -52,7 +52,15 @@ class LabSize(str, Enum):
 
 
 class LabStatus(Enum):
-    """Possible states the user's lab may be in."""
+    """Possible states the user's lab may be in.
+
+    This is not directly equivalent to pod phases. It is instead intended to
```
athornton, line 57: Thanks.  Exactly the sort of commentary I was hoping for.

PR #208, opened 2023-11-01
DM-41188: Rework state management for file servers

rra:
Use the same pattern as the new state management in LabManager: a dictionary of state objects per user, where the user is never removed once we've seen them. In this case, the state object only needs to hold a lock, not a full-blown manager class.

Rather than spawning a separate background task (and Kubernetes watch) for every running file server, use a single background task that watches for all pod changes in the namespace. Use that to catch pod phase changes. Add a background reconcile task to look for any other problems (such as Kubernetes object deletion) and clean up after those.

Add Slack reporting of errors in file server operations.

### DIFF
--- a/src/jupyterlabcontroller/services/fileserver.py
+++ b/src/jupyterlabcontroller/services/fileserver.py
@@ -3,17 +3,38 @@
 from __future__ import annotations
 
 import asyncio
-from collections import defaultdict
+import contextlib
+from dataclasses import dataclass, field
 from datetime import timedelta
 
+from aiojobs import Scheduler
+from safir.datetime import current_datetime
+from safir.slack.blockkit import SlackException
+from safir.slack.webhook import SlackWebhookClient
 from structlog.stdlib import BoundLogger
 
 from ..config import FileserverConfig
-from ..exceptions import MissingObjectError, NotConfiguredError
-from ..models.domain.fileserver import FileserverUserMap
+from ..constants import FILE_SERVER_REFRESH_INTERVAL
+from ..exceptions import (
+    MissingObjectError,
+    NotConfiguredError,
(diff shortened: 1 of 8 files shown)


src/jupyterlabcontroller/services/fileserver.py
```diff
@@ -46,142 +67,261 @@ def __init__(
...
-                    f"Fileserver creation for {username} failed, deleting"
-                    " fileserver objects"
-                )
-                self._logger.exception(msg)
-                await self.delete(username)
+                await self._create_file_server(user)
+            except Exception as e:
+                logger.exception("File server creation failed")
+                await self._maybe_post_slack_exception(e, user.username)
+                logger.info("Cleaning up left-over file server objects")
```
athornton, line 118: Because I can't decide whether it's "left over", "left-over", or "leftover", maybe replace with "orphaned" ?

src/jupyterlabcontroller/storage/kubernetes/fileserver.py
```diff
@@ -199,27 +200,33 @@ async def read_fileserver_state(
...
         return state
 
-    async def wait_for_pod_exit(self, name: str, namespace: str) -> None:
-        """Wait for the fileserver pod spawned by the job to exit.
+    async def watch_pods(self, namespace: str) -> AsyncIterator[PodChange]:
+        """Watches the file server namespace for pod phase changes.
+
+        Technically, this iterator detects any change to a pod and returns its
+        current phase. The change may not be a phase change. That's good
+        enough four our purposes.
```
athornton, line 208: "for"

athornton (approved):
Looks fine.  Couple of minor comment-wording things.

PR #362, opened 2024-03-21
Update JupyterHub to 4.1.0

athornton:


rra (approved):
Needs a change log entry, particularly since we know it breaks our clients with the older handling of XSRF. Otherwise, looks good.

PR #366, opened 2024-03-27
DM-43444: Add more file server routes

rra:
As requested by the Portal team, add a new route for a user or a service acting on behalf of a user to get the status of their file server (/nublado/fileserver/v1/user-status). Return 404 if not running and 200 if running so that clients can ignore the body if they choose. (Currently there is no more-elaborate status to return.)

To maintain parallelism, add an admin .../v1/users/<username> route that returns the same information. Move the DELETE route for admin deletion of a file server under /users to align with the GET route that lists all users and to follow the normal REST URL pattern.


athornton (approved):
Sorry about that.  Thought I had approved it.

PR #372, opened 2024-04-06
tickets/DM-43720: freshen JupyterHub

athornton:
Freshen JupyterHub again, this time with the 4.1.5 release, because their XSRF flailing keeps breaking Firefly.


changelog.d/20240406_123538_athornton_DM_43720.md
```diff
@@ -0,0 +1,6 @@
+<!-- Delete the sections that don't apply -->
+
+### Backwards-incompatible changes
+
+Go to Jupyterhub 4.1.5, which tightens XSRF handling again.
```
rra, line 5: Needs to be a list. I think unless this breaks mobu we should mark this as a bug-fix change instead of backwards-incompatible, since at present I don't think we have any reason to believe upgrading the hub (as opposed to the lab) will break anything.

rra (approved):

PR #379, opened 2024-06-03
tickets/DM-44654: Move to Jupyterhub 5.0.0

athornton:
Also update dependencies.


noxfile.py
```diff
@@ -20,33 +20,42 @@
...
+#
+# In any event, it does no harm to make the fact that we're installing local
+# versions of these modules explicit.
+
 # pip-installable dependencies for development and documentation. This is not
 # used for pytest and typing, since it merges the controller, authenticator,
 # spawner, and inithome dependencies.
 PIP_DEPENDENCIES = [
-    ("--upgrade", "pip", "setuptools", "wheel"),
+    ("--upgrade", "uv", "pip", "setuptools", "wheel"),
```
rra, line 39: You can remove `pip`, `setuptools`, and `wheel` now that we're using `uv`. That will make this go faster.

rra (commented):
Don't we need to make changes for this upstream change?
- New `Authenticator.allow_all` and `allow_existing_users` configuration options. Implicitly allowing all authenticated users when no explicit allow config is provided is no longer the default.

I assume you've checked the spawner form and there are no problems due to this:

- bootstrap is upgraded to 5.3, which may require upgrading if you have custom page templates or use `Spawner.options_form`

athornton:
I thought we'd have to do something, but it still spawns just fine, and I can shut down Labs just fine from the lab side.

athornton:
The spawner form looks a little different but seems to behave just the same.

athornton:
I think the authenticator isn't hurting us because we use a custom authenticator class?

athornton:
Ah.  I have more work to do.  Namely, upgrading Jupyterhub as used by the authenticator!  I missed that.

athornton:
I think it's ready now.  I deleted everyone from users and came back in (after adding allow_all in the authenticator).  It's fine.

athornton:
Don't merge until after summit window 05 June 2024.

CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ Find changes for the upcoming release in the project's [changelog.d directory](h
...
 <!-- scriv-insert-here -->
 
+## 6.1.0 (2024-06-05)
+
+### Other changes
+
+- Jupyterhub 5.0.0 adopted as underlying Hub implementation.
+
+- Move from `pip` to `uv` in `nox` and CI.
+
```
rra, line 16: So, the idea behind how scriv is supposed to work is that we leave the fragments as fragments on all the feature branches. Then, when we're ready to make a release, we make a *new* branch, run `scriv collect` there (and any other release tasks like `make update` / `nox -s update-deps`), and then merge that right before the release. That release prep branch doesn't require review.

No need to move these back into fragments to redo this, just a note for next time, but note that there's now another change going into this release so now `scriv collect` won't work in that other branch, which is why `scriv collect` should be the last thing before a release on its own branch with only release stuff.

authenticator/pyproject.toml
```diff
@@ -20,7 +20,7 @@ classifiers = [
 ]
 requires-python = ">=3.10"
 dependencies = [
-    "jupyterhub<5",
+    "jupyterhub==5",
```
rra, line 23: I don't think you want the plugin to only allow version 5.0.0 exactly, which is what that does. Also, the plugin should continue to work fine with earlier versions of jupyterhub, so I think leaving this at `jupyterhub<6` is fine until we know for sure that we did something that broke compatibility.

spawner/pyproject.toml
```diff
@@ -22,7 +22,7 @@ requires-python = ">=3.10"
 dependencies = [
     "httpx<0.26",
     "httpx-sse<0.4",
-    "jupyterhub<5",
+    "jupyterhub==5",
```
rra, line 25: Same comment here about the version number.

rra (approved):
Looks good once the version pins for `authenticator` and `spawner` are fixed.

CHANGELOG.md
athornton: Easy enough to revert, though.

authenticator/pyproject.toml
athornton: Huh.  I thought that meant "Anything with major version 5"

PR #380, opened 2024-06-05
tickets/DM-44689: Add resource limits/requests to prepulled pods

athornton:
Add limits and requests to prepulled pods.

### DIFF
--- a/controller/src/controller/services/builder/prepuller.py
+++ b/controller/src/controller/services/builder/prepuller.py
@@ -68,8 +68,8 @@ class PrepullerBuilder:
                         image=image.reference_with_digest,
                         resources=V1ResourceRequirements(
                             limits={
-                                "cpu": "1",
-                                "memory": "512Mi",
+                                "cpu": "1m",
+                                "memory": "16Mi",
                             },
                             requests={
                                 "cpu": "1m",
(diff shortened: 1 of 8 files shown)


athornton:
Rebase when DM-44654 merged.  Also do scriv collect at that point.

controller/src/controller/services/builder/prepuller.py
```diff
@@ -65,6 +66,16 @@ def build_pod(self, image: RSPImage, node: str) -> V1Pod:
...
+                        resources=V1ResourceRequirements(
+                            limits={
+                                "cpu": "1",
+                                "memory": "512Mi",
+                            },
+                            requests={
+                                "cpu": "1m",
+                                "memory": "16Mi",
+                            },
+                        ),
```
rra, line 78: Let's set the limits to the same as the requests. Running `/bin/true` should never take that much memory.

rra (approved):
Looks good except let's make the limits smaller.

controller/src/controller/services/builder/prepuller.py
rra: Also if you remove the trailing commas, ruff should  manage to make this a lot more compact.

PR #388, opened 2024-08-15
DM-45794: Catch another node image variation

rra:
The node information from Kubernetes apparently can have a None value for node.status.images[].names instead of a list. Catch that case and discard that image record, since we can't do anything with it.


athornton (approved):
I am concerned in that I don't know how names could be empty.  But less concerned since it is an optional field in the spec, so the K8s implementation is within its rights to return None.

PR #395, opened 2024-08-16
tickets/DM-45805: make `/tmp` come from memory

athornton:
And cap size of `/tmp` usage.


rra (commented):
Code seems fine. Can we add a new configuration option that controls this? `config.lab.memoryTmp` defaulting to `true`, maybe? It feels highly likely to be something different sites will want to have different policies for. Or maybe it would be better to have `config.lab.tmpSource` be an enum with values `memory` and `node-local`? Something like that.

I don't think the ratio of memory to local cap needs to be configurable, but it would be good to pull it into constants.

If you spawn a lab with this configuration, does `/tmp` show up as being tmpfs in `df`, etc? Just curious.

athornton:
It does show up as tmpfs.  I will add a configurable option.

```
Filesystem                   1K-blocks      Used  Available Use% Mounted on
overlay                      305267820  54031828  251219608  18% /
tmpfs                            65536         0      65536   0% /dev
tmpfs                          2097152    518112    1579040  25% /tmp
```

athornton:
How about just making emptyDir an explicit volume type like we've done with HostPath and PVC and NFS ?

rra:
I thought about that but the implication is that we would then add support for mounting some arbitrary volume type on `/tmp`, and while that's not an unreasonable thing to support, we have no requirement for it and I'm not sure we should bother.

It's awkward; I'm not sure the best way to represent it, particularly since I don't want people to specify an explicit memory limit in the configuration since it should vary by image type. A boolean is probably the simplest thing, and we can always change it later.

athornton:
I went with the two-valued enum.

changelog.d/20240816_093119_athornton_DM_45805.md
```diff
@@ -0,0 +1,5 @@
+<!-- Delete the sections that don't apply -->
+
+### Other changes
```
rra, line 3: Since we changed the default, I think this is a backwards-incompatible change (it will require attention from people running Nublado if they don't want `/tmp` to be a memory file system). The next release would therefore be 7.0.0.

controller/src/controller/services/builder/lab.py
```diff
@@ -670,10 +678,24 @@ def _build_pod_env_volume(self, username: str) -> MountedVolume:
...
-    def _build_pod_tmp_volume(self) -> MountedVolume:
-        """Build the volume that provides a writable tmpfs :file:`/tmp`."""
+    def _build_pod_tmp_volume(self, mem_size: int) -> MountedVolume:
+        """Build the volume that provides a writable tmpfs :file:`/tmp`.
+
+        Note that it is the `Memory` medium setting that forces this to
+        be tmpfs rather than ephemeral storage, and therefore usage will
+        come from the pod memory allocation rather than disk space.
+        """
+        medium: str | None = None
```
rra, line 688: This should work fine without the type annotation. mypy recognizes this construction.
```suggestion
        medium = None
```

rra (approved):
Looks good, just a couple of minor nits.

PR #401, opened 2024-09-13
tickets/DM-46157: Move Nublado Client into Nublado monorepo

athornton:
This combines the features of the somewhat-diverged Mobu and Noteburst clients, and moves the whole thing into the Nublado monorepo.

Ghostwriter is the only current customer, and I imagine we will find some bugs as we rework Mobu and Noteburst to use the unified client.  More with Noteburst, most likely, since this is very close to the Mobu client.


docs/dev/client.rst
```diff
@@ -0,0 +1,79 @@
+############
```
jonathansick, line 1: This is a user guide for the Nublado client right? In that case, it doesn't belong in the Nublado developer documentation, but instead as a separate part of the Nublado docs.

docs/dev/client.rst
```diff
@@ -0,0 +1,79 @@
...
+
+#. ``run_python()``.  This runs a string representing arbitrary Python code and returns streamed results (which is to say, ``stdout`` wihin each cell).  Of course, this code will run in the Lab environment specified by the session, which will be bound to a particuar kernel (usually ``LSST`` within the RSP), and will have access to whatever is installed in that kernel.
+#. ``run_notebook()``.  This runs each cell of a supplied notebook using ``run_python()``, accumulating the results and ultimately returning them as a list of cell outputs.
+#.  ``run_notebook_via_rsp_extension()``.  As is obvious from the name, this is not a standard Jupyter feature.  This runs a notebook in the same way ``Times-Square`` does, by using a Rubin Observatory-specific extension within the user lab that, in turn, uses ``nbconvert`` to execute notebooks and return their rendered form.  If you need to execute a notebook and capture output that did not go to stdout (for instance, the Javascript created by a Bokeh call, that will ultimately run in your browser), this is at present the way to do it.
```
jonathansick, line 44:
```suggestion
#.  ``run_notebook_via_rsp_extension()``.  This executes a notebook via the ``nbexec`` RSP JupyterLab extension.  `Times Square <https://times-square.lsst.io>`__ and `Noteburst <https://noteburst.lsst.io>`__ use this method. The ``nbexec`` JupyterLab extensions runs within the user lab, and in turn, uses ``nbconvert`` to execute notebooks and return their rendered form.  If you need to execute a notebook and capture output that did not go to stdout (for instance, the Javascript created by a Bokeh call, that will ultimately run in your browser), this is at present the way to do it.
```

docs/dev/client.rst
```diff
@@ -0,0 +1,79 @@
...
+In the client test suite, there is a context manager ``jupyter`` which is an illustration of how to create a MockJupyter (which is fundamentally based on the ``respx`` class for testing ``httpx`` services) and monkeypatch in a websocket emulator.
+Once you have a mocked Jupyter in play, you should not need to do very much else: use a regular ``NubladoClient``, and if you have configured it and the mock for the same base URL, the mock should respond appropriately.
+
+Mocking Payloads
+----------------
+
+The attentive reader will have noticed a potential problem.
+The Python code being used as a client payload is expected, in the wild, to run within an RSP kernel.
+The ``LSST`` kernel is extremely heavyweight and has all kinds of features not found in a vanilla Python installation.
+How can we unit-test this without installing the DM stack in our test suite?
```
jonathansick, line 64: Consider dropping this paragraph or cutting to the chase.

docs/dev/client.rst
```diff
@@ -0,0 +1,79 @@
+############
+Client guide
+############
+
+This page describes the use of the ``NubladoClient`` class and especially its provided testing classes.
```
jonathansick, line 5: These docs should mention how to install the client from PyPI and also mention the full Python namespace early rather than just ``NubladoClient``. If we can make API docs for the NubladoClient and its exceptions and models, even better.

docs/dev/client.rst
```diff
@@ -0,0 +1,79 @@
...
+#. Authenticate to the Hub with the ``auth_to_hub()`` method.
+#. Determine whether or not you have a running lab with ``is_lab_stopped()``.
+#. If you need to, spawn a lab with ``spawn_lab()``.
+#. Wait for the lab to spawn by looping through ``watch_spawn_progress()`` until you get a progress message indicating the lab is ready.
+#. Authenticate to the Lab with ``auth_to_lab()``
+#. Use ``open_lab_session()`` as a context manager to interact with your lab.
+#. Do whatever it is you wanted to do with the lab; more below.
+#. When done, use ``stop_lab()`` to shut down the lab, if desired.
+
+The ``client_test`` test in the client test suite steps through this process and may be a useful model.
```
jonathansick, line 33: Try dropping an example into the docs, because finding the `client_test` module? function? is a little inconvenient.

docs/dev/client.rst
```diff
@@ -0,0 +1,79 @@
...
+
+.. _lab-interaction:
+
+Interacting With The Lab
+========================
+
+``NubladoClient`` provides three methods of interacting with a spawned lab.  These are methods on the ``JupyterLabSession`` object you will have available inside the session context manager.  They are:
```
jonathansick, line 40: I think orienting the docs around use cases and including examples would be good.

client/pyproject.toml
```diff
@@ -0,0 +1,114 @@
...
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.12",
+    "Intended Audience :: Developers",
+    "Natural Language :: English",
+    "Operating System :: POSIX",
+    "Typing :: Typed",
+]
+requires-python = ">=3.12"
+# Use requirements/main.in for runtime dependencies instead.
+dependencies = []
```
jonathansick, line 22: Ah, this doesn't work for Python packages, only for our Docker-based apps. You'll need to put dependencies into the `pyproject.toml` file.

jonathansick: Theoretically dev.in/.txt are fine with Python packages, but there's no point having a main.in/.txt file.

client/src/rubin/nublado/client/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .nubladoclient import NubladoClient
+
+__all__ = ["NubladoClient"]
```
jonathansick, line 3: There are a lot of modules in the Nublado client. I wonder if it's be more user friendly to make some of these modules private and import their APIs into a higher level like or, or at most have a `rubin.nublado.client`, `rubin.nublado.client.exceptions` and `rubin.nublado.client.models` spaces in the public API?

client/src/rubin/nublado/client/exceptions.py
```diff
@@ -0,0 +1,456 @@
...
+
+    Returns
+    -------
+    str
+        Sanitized string.
+    """
+    return _ANSI_REGEX.sub("", string)
+
+
+class MobuSlackException(SlackException):
```
jonathansick, line 60: Will it be confusing to have the "Mobu" naming in the Nublado client?

jonathansick (changes requested):
Most of the comments are ideas/suggestions, but the critical one that needs to be fixed is putting the dependencies in pyproject.toml and dropping the `requirements/main.txt` file.

client/pyproject.toml
```diff
@@ -0,0 +1,114 @@
...
+dynamic = ["version"]
+
+[project.scripts]
+rsp_client = "rubin.nublado.client.cli:main"
```
jonathansick, line 26: I don't this this `cli` module exists? So this'll need to be dropped in the first release.

client/src/rubin/nublado/client/exceptions.py
athornton: It will.  That was a transplantation oversight.

client/pyproject.toml
athornton: Oh, yeah, that was something I never finished.  Do you think it's a better idea to:

1) have a CLI class, and have a nublado client that you associate a username, token, and base URL with (or alternatively, token and base URL and ask Gafaelfawr for the name), and then have a CLI client, or
2) Assume that if you're using this class, you're happy with it being a Python object and not having an immediate CLI beyond the REPL.

I had started at 1, but never actually wrote the CLI tool, and after writing a bunch of hooks in Ghostwriter, find I don't miss it.

client/src/rubin/nublado/client/__init__.py
athornton: I think in general a simplified hierarchy exposed to the user is a good idea, although I'll need to think about how to structure that.  We also want `.testing.MockJupyter` exposed, probably there, because service authors are going to want that piece.  But all the ancillary stuff in `.testing.jupyter` probably should stay hidden.  Maybe lifting the mock into `.testing`, the exception from the extension model into `.exceptions`. and leaving the models pretty much alone is the right thing?

Anyway, yes, just need to think about exactly how.

docs/dev/client.rst
athornton: OK.  To what degree is it legitimate to point at Ghostwriter?  Because the Ghostwriter hooks are (I think) a pretty good example of some things you might want to do to exercise `run_python()`

docs/dev/client.rst
athornton: Noted (it's not `nbexec` anymore, just the `/rubin/execution` endpoint of the server extension, but I'll call it out).

docs/dev/client.rst
athornton: OK, I'll hoist it up a level.

client/pyproject.toml
athornton: Done.

client/src/rubin/nublado/client/__init__.py
athornton: Done.  Not sure if it's much better, but import names are slightly shorter, anyway.

client/pyproject.toml
athornton: Removed for now.  I'd be willing to listen to an argument for 1, but I'm not sure there's any reason for anyone, even me, to have a command-line tool to run code in a particular RSP as a particular user.  Noteburst if you love terminal windows, maybe?

client/src/rubin/nublado/client/exceptions.py
athornton: Fixed.

docs/dev/client.rst
athornton: I do not know how to make API docs, or at least not how to autogenerate them.

docs/dev/client.rst
athornton: Done.

docs/client/index.rst
```diff
@@ -0,0 +1,204 @@
...
+
+.. _Mocks:
+
+Mocks
+=====
+
+In the module ``rubin.nublado.client.testing`` you will find the ``MockJupyter`` class.
+This provides a simulation of the RSP Nublado Hub/Proxy/Controller environment, as well as a partial simulation of the Labs it spawns.
+The reason you would use this is to be able to meaningfully test your service without having to test against a live RSP or spin up your own RSP to test the service against.
+Although there are quite a few additional classes within the module, ``MockJupyter`` should be the only one you need directly, except to set up the test fixure.
```
stvoutsin, line 110:
```suggestion
Although there are quite a few additional classes within the module, ``MockJupyter`` should be the only one you need directly, except to set up the test fixture.
```

docs/client/index.rst
```diff
@@ -0,0 +1,204 @@
...
+The ``client_test`` test in the client test suite steps through this process and may be a useful model.
+
+.. _lab-interaction:
+
+Interacting With The Lab
+========================
+
+``NubladoClient`` provides three methods of interacting with a spawned lab.  These are methods on the ``JupyterLabSession`` object you will have available inside the session context manager.  They are:
+
+#. ``run_python()``.  This runs a string representing arbitrary Python code and returns streamed results (which is to say, ``stdout`` wihin each cell).  Of course, this code will run in the Lab environment specified by the session, which will be bound to a particuar kernel (usually ``LSST`` within the RSP), and will have access to whatever is installed in that kernel.
```
stvoutsin, line 51:
```suggestion
#. ``run_python()``.  This runs a string representing arbitrary Python code and returns streamed results (which is to say, ``stdout`` within each cell).  Of course, this code will run in the Lab environment specified by the session, which will be bound to a particular kernel (usually ``LSST`` within the RSP), and will have access to whatever is installed in that kernel.
```

docs/client/index.rst
```diff
@@ -0,0 +1,204 @@
...
+.. prompt:: bash
+
+    pip install rubin-nublado-client
+
+.. _client-usage:
+
+Using the Client
+================
+
+The ``NubladoClient`` is designed to make interaction with JuptyerHub and Jupyterlab, as they are configured in the RSP environment, easy.
```
stvoutsin, line 21:
```suggestion
The ``NubladoClient`` is designed to make interaction with JupyterHub and Jupyterlab, as they are configured in the RSP environment, easy.
```

client/src/rubin/nublado/client/nubladoclient.py
```diff
@@ -0,0 +1,1067 @@
...
+from .models import NotebookExecutionResult, NubladoImage, User
+
+P = ParamSpec("P")
+T = TypeVar("T")
+
+__all__ = ["NubladoClient", "JupyterLabSession"]
+
+
+@dataclass(frozen=True, slots=True)
+class SpawnProgressMessage:
```
stvoutsin, line 51: Should this perhaps be in models?

client/src/rubin/nublado/client/nubladoclient.py
```diff
@@ -0,0 +1,1067 @@
...
+            now = current_datetime(microseconds=True)
+            elapsed = int((now - self._start).total_seconds())
+            status = "complete" if event.ready else "in progress"
+            msg = f"Spawn {status} ({elapsed}s elapsed): {event.message}"
+            self._logger.info(msg, elapsed=elapsed, status=status)
+            yield event
+
+
+@dataclass(frozen=True, slots=True)
+class JupyterOutput:
```
stvoutsin, line 121: Maybe this can go in models as well?

docs/client/index.rst
```diff
@@ -0,0 +1,204 @@
...
+    async def ensure_lab() -> None:
+        """Start a Lab if one is not present."""
+        client = NubladoClient(
+ 	    user=User(username="some-user", token="some-token"),
+            base_url="https://data.example.org"
+        )
+        await client.auth_to_hub()
+        stopped = await client.is_lab_stopped()
+	if stopped:
+	    image=NubladoImageByClass(
```
stvoutsin, line 86: Looks like some tabs that need to be spaces here

stvoutsin (changes requested):

client/src/rubin/nublado/client/exceptions.py
```diff
@@ -0,0 +1,448 @@
...
+"""Regex that matches ANSI escape sequences."""
+
+__all__ = [
+    "CodeExecutionError",
+    "JupyterProtocolError",
+    "JupyterTimeoutError",
+    "JupyterWebError",
+    "JupyterWebSocketError",
+    "NubladoClientSlackException",
+    "NubladoClientSlackWebException",
```
stvoutsin, line 33:
```suggestion
    "NubladoClientSlackWebException",
    "ExecutionAPIError"
```

client/src/rubin/nublado/client/models/_extension.py
```diff
@@ -0,0 +1,42 @@
+"""Models for notebook execution via the ``/user/:username/rubin/execute``
```
stvoutsin, line 1: I'm guessing this is probably correct and I'm missing some context, but are these two different endpoints:
```
/rubin/execution
/rubin/execute
```

docs/dev/client.rst
athornton: Done.

docs/dev/client.rst
athornton: Docs restructured and improved.  No API autogen though.

docs/dev/client.rst
athornton: Eh, just wrote some long-form examples in the docs.

client/src/rubin/nublado/client/models/_extension.py
athornton: Should be `/execution`

client/src/rubin/nublado/client/_util.py
```diff
@@ -61,35 +69,101 @@ def normalize_source_by_cell(notebook: str) -> list[str]:
...
-        )
+        x["source"]
+        for x in obj["cells"]
+        if x["cell_type"] == "code" and "source" in x and x["source"]
     ]
 
 
-def extract_source_by_cell(notebook: str) -> list[list[str]]:
-    """Extract all non-empty "code" cells' "source" lines as a list of strings.
+def source_list_by_cell(notebook: str) -> list[list[str]]:
```
stvoutsin, line 74: Would it make sense to merge `source_list_by_cell` & `source_string_by_cell` into one method, which will take the `NotebookForm` as a parameter and call out the according method (notebook_to_x.. ) ?

stvoutsin (approved):

client/src/rubin/nublado/client/_util.py
athornton: Oh, I should just get rid of NotebookForm.  Since I'm using the file API instead of the contents one I don't need it.

athornton:
Except I hadn't pushed that pyproject.toml update.

PR #410, opened 2024-09-27
tickets/DM-46538: Add jupyterlab-base build to nublado monorepo

athornton:


athornton:
This adds the UI side of the two-stack sciplat lab build to nublado as jupyterlab-base.

afausti (approved):
LGTM Adam kindly went over the build process with me so I understand it a little better. The separation of concerns in the base and sciplatlab build steps makes a lot of sense and will make it easier to extend the RSP to other stacks in the future.

PR #428, opened 2024-11-15
Reset hub_xsrf on each hub_login

athornton:


athornton:
Nope, the naive approach ain't gonna do it.

client/src/rubin/nublado/client/nubladoclient.py
```diff
@@ -807,22 +807,26 @@ def __init__(
...
         self._logger = logger
-
+        self._timeout = timeout
+        self._base_url = base_url
+        self._initialize_client()
         # Construct a connection pool to use for requests to JupyterHub. We
         # have to create a separate connection pool for every user, since
         # each will get user-specific cookies set by JupyterHub. If we shared
         # connection pools, users would overwrite each other's cookies and
         # get authentication failures from labs.
```
rra, line 817: This comment is now misplaced and should either be a docstring on `_initialize_client` or before the invocation of `_initialize_client`.

client/src/rubin/nublado/client/nubladoclient.py
```diff
@@ -856,6 +860,10 @@ async def auth_to_hub(self) -> None:
         JupyterProtocolError
             Raised if no ``_xsrf`` cookie was set in the reply from the lab.
         """
+        # Remove hub xsrf token.
+        self._hub_xsrf = None
+        # Reinitialize HTTP client
```
rra, line 865: The comment, or somewhere, should say why this is needed so that we can remember all of this debugging. I'd probably add a note section to the docstring with a pretty comprehensive explanation of what we figured out.

client/tests/conftest.py
```diff
@@ -52,9 +53,20 @@ def configured_logger() -> BoundLogger:
     return structlog.get_logger("nublado-client")
 
 
+def _create_mock_token(username: str, token: str) -> str:
+    # A mock token is: "gt-<base-64 encoded username>.<base64 encoded token>"
+    # That is then decoded to extract the username in the Jupyter mock.
+    enc_u = urlsafe_b64encode(username.encode()).decode("utf-8")
+    enc_t = urlsafe_b64encode(token.encode()).decode("utf-8")
```
rra, line 60:
```suggestion
    enc_u = urlsafe_b64encode(username.encode()).decode()
    enc_t = urlsafe_b64encode(token.encode()).decode()
```
UTF-8 is the default.

rra (approved):
I think it needs more comments to capture why we need to do this, but otherwise looks good. It's not clear from the pull request or commit message why the test suite change is needed; ideally that should be in the commit message for later understanding of the commit.

PR #430, opened 2024-11-26
DM-47811: Fixes to address FAST001 and FAST002 ruff rules and update-deps

stvoutsin:
**Summary**

Fix ruff rules [FAST001](https://docs.astral.sh/ruff/rules/fast-api-redundant-response-model/) & [FAST002](https://docs.astral.sh/ruff/rules/fast-api-non-annotated-dependency/) which were introduced in ruff version [0.8.0](https://github.com/astral-sh/ruff/releases/tag/0.8.0)

This PR addresses this in the following way:

- Updates the dependencies including ruff
- Removes the response_model param which is no longer needed
- Includes autofixed import ordering changes


changelog.d/20241126_232616_steliosvoutsinas.md
rra, line 1: No user-visible changes so no change log needed.

rra (approved):
Looks great apart from the change log. Thanks!

PR #432, opened 2024-12-10
update hub for newer upstream (+helm chart)

athornton:


hub/page.html
```diff
@@ -0,0 +1,30 @@
+{% extends "templates/page.html" %}
```
stvoutsin, line 1: The only very minor suggestion I have is, if you are expecting to be adding more templates in the future, perhaps I'd place under a folder templates/ rather than directly under hub just to keep the top directory clean.

PR #435, opened 2024-12-12
DM-48124: Update dependencies and build system

rra:
Convert all of the Docker builds to use uv to install packages, which allows removing the update-deps-no-hashes nox session and all of its supporting code. uv understands how to verify hashes if present and skip the verification if the hashes aren't present.

Encode the startup command of the controller directly in the Dockerfile and remove the one-line shell script that was previously used to start it.

Update to the latest Ruff and shared Ruff configuration file. Move Nublado-specific overrides out of the shared Ruff configuration file into `pyproject.toml` where project-specific overrides should live. Fix a few issues uncovered by Ruff that can be fixed instead of suppressed.

Remove dead configuration from `client/pyproject.toml` and add some additional metadata. Remove unnecessary dependencies.

Remove the now-unused `Dockerfile.sciplat-lab`. That build was moved to a different repository.

Update pre-commit and Python dependencies.


athornton (approved):
LGTM.

PR #459, opened 2025-01-23
DM-48546: Add support for blocking spawn with a quota flag

rra:
Gafaelfawr now supports a flag in the user's notebook quota saying whether that user should be allowed to spawn a notebook. Implement support for that flag in the Nublado controller. If it is set to false, return an error message instead of a spawn form, and reject the actual spawn with a 403 error.


athornton (approved):
LGTM

PR #460, opened 2025-01-24
DM-48576: Suppress some client warnings for async generators

rra:
Python 3.13 adds a new warning when the `aclose` method of an async generator is not called explicitly. Async generators should ideally be explicitly shut down with `aclose` when they are no longer in use. Fix typing in some cases where async generators were returned but typed as async iterators (which don't require `aclose`), and close more generators explicitly. Include some async iterators from third party libraries that are actually generators, which requires working around the typing and some future-proofing for API changes.

Ignore the remaining warning that's internal to httpx-sse and can't be fixed at the client level.

### DIFF
--- a/client/src/rubin/nublado/client/nubladoclient.py
+++ b/client/src/rubin/nublado/client/nubladoclient.py
@@ -8,7 +8,8 @@ from __future__ import annotations
 
 import asyncio
 import json
-from collections.abc import AsyncIterator, Callable, Coroutine
+from collections.abc import AsyncGenerator, Callable, Coroutine
+from contextlib import AbstractAsyncContextManager, aclosing, suppress
 from datetime import UTC, datetime, timedelta
 from functools import wraps
 from pathlib import Path
@@ -68,7 +69,7 @@ class JupyterSpawnProgress:
         self._logger = logger
         self._start = datetime.now(tz=UTC)
 
-    async def __aiter__(self) -> AsyncIterator[SpawnProgressMessage]:
+    async def __aiter__(self) -> AsyncGenerator[SpawnProgressMessage, None]:
         """Iterate over spawn progress events.
 
         Yields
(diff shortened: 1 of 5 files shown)


client/tests/client/client_test.py
```diff
@@ -37,12 +38,13 @@ async def test_hub_flow(
...
     progress = configured_client.watch_spawn_progress()
     progress_pct = -1
-    async with asyncio.timeout(30):
-        async for message in progress:
-            if message.ready:
-                break
-            assert message.progress > progress_pct
-            progress_pct = message.progress
+    async with aclosing(progress):
+        async with asyncio.timeout(30):
```
fajpunk, line 42: Just curious if the `aclosing` has to be outside the `timeout`, or if it doesn't matter.

client/src/rubin/nublado/client/nubladoclient.py
```diff
@@ -82,27 +83,36 @@ async def __aiter__(self) -> AsyncIterator[SpawnProgressMessage]:
...
+                status = "complete" if event.ready else "in progress"
+                msg = f"Spawn {status} ({elapsed}s elapsed): {event.message}"
+                self._logger.info(msg, elapsed=elapsed, status=status)
+                yield event
+        finally:
+            # aiter_sse actually returns an asynchronous generator, which
+            # therefore is supposed to be explicitly closed with alcose()
+            # after break, unlike a true iterator. Handle this case to suppress
+            # warnings in Python 3.13.
+            with suppress(AttributeError):
```
fajpunk, line 114: Why do we have to handle `AttributeError`s here (and elsewhere), instead of using `aclosing`?

fajpunk (approved):

client/tests/client/client_test.py
rra: It shouldn't matter. It will close the generator in the `finally` block when exited by the exception raised by `asyncio.timeout` either way.

client/src/rubin/nublado/client/nubladoclient.py
rra: This is because the upstream typing says that they're returning an async iterator, but it's actually an async generator. Async iterators don't have `aclose` methods, and they could replace the generator with an iterator at any time and not break the interface guarantee. So we have to allow for the possibility that `aclose` doesn't exist, because it really is an iterator.

I've now refactored this into a helper function (which we could lift into Safir if we wanted to pursue this elsewhere) and added a bunch more documentation to make this clearer.

fajpunk: Ohhhhhhh that's to suppress the case where `aclose` isn't there. I mistakenly thought it was to catch an `AttributeError` from the contents of `aclose`.  This makes total sense.

rra: Oh, yeah, the suppress might be a bit too broadly scoped. I didn't think about that. I suppose I should actually use `getattr` just to be safer?

PR #465, opened 2025-02-01
DM-48696: Add metrics events

rra:
Add metrics events for spawn successes, spawn failures, and the count of active, running labs. The spawn success and failure events include the username, Docker image reference, CPU limit, memory limit, and elapsed time. We may want to add more fields later.

The active lab count is logged via the background reconcile task.

### DIFF
--- a/controller/src/controller/events.py
+++ b/controller/src/controller/events.py
@@ -1,0 +1,114 @@
+"""Metrics events for the Nublado controller."""
+
+from __future__ import annotations
+
+from datetime import timedelta
+
+from pydantic import Field
+from safir.dependencies.metrics import EventMaker
+from safir.metrics import EventManager, EventPayload
+
+__all__ = [
+    "ActiveLabsEvent",
+    "LabEvents",
+    "LabMetadata",
+    "SpawnFailureEvent",
+    "SpawnSuccessEvent",
+]
+
+
+class ActiveLabsEvent(EventPayload):
+    """Current count of the number of active labs.
+
+    Notes
+    -----
+    This is really a gauge metric that is measured periodically, not an event.
+    For now, the Nublado controller uses the event system to log this metric,
+    since that's the system we have in place. If we later have a proper
+    metrics system for storing measurements, this event should move to that
+    system.
+    """
+
+    count: int = Field(
+        ...,
+        title="Active labs",
+        description="Number of currently-running labs",
+    )
+
+
+class LabMetadata(EventPayload):
+    """Common lab metadata for events."""
+
+    image: str = Field(
+        ...,
+        title="Lab image",
+        description="Docker reference for the lab image",
+    )
+
+    cpu_limit: float = Field(
+        ...,
+        title="Lab CPU limit",
+        description="Kubernetes pod limit of CPU equivalents",
+    )
+
+    memory_limit: int = Field(
+        ...,
+        title="Lab memory limit",
+        description="Kubernetes pod limit of memory in bytes",
+    )
+
+
+class SpawnFailureEvent(LabMetadata):
+    """A lab spawn failed."""
+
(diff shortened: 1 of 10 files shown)


controller/src/controller/services/lab.py
```diff
@@ -637,12 +666,19 @@ async def reconcile(self) -> None:
...
+        # as long as the user hasn't started some other operation in the
+        # meantime.
         await self._delete_completed_labs()
 
+        # Finally, count the active labs and log a metric.
+        count = 0
+        for state in self._labs.values():
+            if state.state and state.state.status == LabStatus.RUNNING:
+                count += 1
+        await self._events.active.publish(ActiveLabsEvent(count=count))
```
fajpunk, line 680: Would we want counts for labs with other statuses? If so, what about adding a `status` field to the event, and folks can filter on whatever they want?

controller/tests/support/config.py
```diff
@@ -36,6 +39,10 @@ async def configure(
     Config
         New configuration.
     """
+    monkeypatch.setenv("METRICS_APPLICATION", "controller")
+    monkeypatch.setenv("METRICS_ENABLED", "false")
+    monkeypatch.setenv("METRICS_MOCK", "true")
+
```
fajpunk, line 45: Could this be in an autouse fixture so you don't have to thread a `monkeypatch` fixture through all the tests?

controller/src/controller/events.py
```diff
@@ -0,0 +1,114 @@
...
+
+    username: str = Field(
+        ..., title="Username", description="User who attempted to spawn a lab"
+    )
+
+    elapsed: timedelta = Field(
+        ...,
+        title="Duration of spawn attempt",
+        description="How long the spawn took before it failed",
+    )
```
fajpunk, line 72: I've been using one event for both successes and failures, with a `success` bool field. I slightly prefer that, though I don't have a _great_argument for it:
* We don't need two different metrics events for ever conceptual event. Here, I see the conceptual event as a spawn, and the success or failure is a property.
* Going along with the 'success or failure as an attribute' argument, I've been trying to put all relevant information about a domain event on a single metrics event, because I think it's easier to query on the other end. It's one count bunch of `WHERE` clauses on one event type vs a bunch of `WHERE` clauses on different event types and adding up counts.

On the other hand, if there's a lot attributes that are only on successful events vs failure events, then it can be _more_ confusing to query because the querier needs to know what's supposed to be on what, vs all of the attributes of a given event being relevant. But so far, most (all?) cases have had the same attributes, with the only difference being success or failure.

I don't feel super strongly, but I think we should probably use the same strategy everywhere if we can.

controller/src/controller/services/lab.py
rra: Why wouldn't we log separate metrics for each type of lab? Is there some benefit to combining them that I'm missing?

controller/src/controller/events.py
rra: The reason why I logged these as separate events is that I think it would always be an error to query this metric without specifying either success == true or success == false. To me, that implies they should be separate metrics to avoid the risk that someone might think it's meaningful to combine successes and failures.

controller/tests/support/config.py
rra: Good call, yes. Fixing, thanks!

controller/src/controller/events.py
fajpunk: After talking it through, I'm sold to keep success and failure as separate events.

controller/src/controller/services/lab.py
fajpunk: Similar to the success/failure case, if it never or rarely makes sense to combine them, then I'm sold on separate events for counts of different statuses.

PR #466, opened 2025-02-03
tickets/DM-48682-reaper: Move rsp-reaper into nublado and put its models into controller.

athornton:
This relocates rsp-reaper into the nublado monorepo and moves its models into nublado controller (since the controller will soon pick up the image policies from the reaper, but that will be a separate PR)

There will be some classmethods to generate MenuImages from RegistryImages or something like that in a future PR, but that too is not here.  Mostly this is just a forklift of rsp-reaper into place.


rra (changes requested):
As discusssed in the meeting, please break this up into at least three separate PRs: the changes to the image tag data structures and their tests, the change to the configuration structure and corresponding code changes, and the addition of the reaper to the repository.

Also, we shouldn't add the reaper with copies of code from the controller to the repository. Part of adding it needs to be refactoring it to either be part of the controller subtree or to move all of the shared code into a library that can be used by both.

PR #467, opened 2025-02-04
DM-48714: Do not query Kubernetes for pod status

rra:
Do not query Kubernetes for pod status when responding to status requests, and instead assume the internal state (which is reconciled periodically) is correct. The constant Kubernetes API requests for `Pod` status seemed to be overwhelming the control plane when there were a lot of pods running. Continue to ask Kubernetes directly immediately before spawn.


athornton (approved):
I hope this turns out to be the right heuristic, but I guess we will find out on Friday.  It seems to me like it should work just fine, assuming that the periodic reconciliation is reasonably frequent and that this does take enough load off the control plane that we get the right status information when we actually need it.

PR #469, opened 2025-02-04
DM-48761: Add tolerations to prepuller pods

rra:
When Nublado was configured to add tolerations to lab pods, those tolerations were not also applied to prepuller pods. This meant that prepuller pods were never run if the node pool targeted by Nublado was tainted, such as when using a node pool restricted to only Nublado labs. Add the tolerations to prepuller pods as well.


athornton (approved):
LGTM

PR #470, opened 2025-02-04
Add date and age fields to RSPTag

athornton:


controller/src/controller/models/domain/rsptag.py
```diff
@@ -106,6 +107,38 @@ class RSPImageType(Enum):
...
+    year = tagdata.get("year")
+    if not year:
+        return None
+    week = tagdata.get("week")
+    if year and week:
+        jan1 = datetime.datetime(int(year), 1, 1, tzinfo=datetime.UTC)
+        # This isn't exact, it's not isoformat, and it's not pretty, but it is
+        # also close enough for our purposes.
+        # We start counting at week one in our tags.
+        return jan1 + datetime.timedelta(weeks=int(week) - 1)
```
rra, line 132: I think this is a complicated way of writing `datetime.fromisocalendar(int(year), int(week), 4)`. You do unfortunately have to specify the day of the week, but all of this is specific to Rubin practices anyway so we should just set whatever day of the week it is that we tag weeklies. The above is Friday; I'm not sure if that's right.

controller/src/controller/models/domain/rsptag.py
```diff
@@ -106,6 +107,38 @@ class RSPImageType(Enum):
 ]
 
 
+def _calculate_date(tagdata: dict[str, str]) -> datetime.datetime | None:
```
rra, line 110: This should be a method rather than a standalone function. It's only useful in the context of the parser in the class.

controller/src/controller/models/domain/rsptag.py
```diff
@@ -130,6 +163,22 @@ class RSPImageTag:
...
 
+    date: datetime.datetime | None = None
+    """When the image was created, or as close as we can get to that.
+    We try to derive this from the tag string: For RSP daily or weekly
+    tags (or experimentals in one of those formats), we can calculate
+    this to within a day or a week, which is good enough for display
+    purposes.  Otherwise, we may be able to extract this info from
+    the registry, but even if we can, it may be image upload time
+    rather than creation time.
+    """
```
rra, line 174: Blank line after the summary sentence in docstrings.

This shouldn't be an optional field. You always set it to something, so you can just declare it as `datetime | None` without the `= None`.That lets you drop the `kw_only` on the child class, although that doesn't really matter for how we use it.

controller/src/controller/models/domain/rsptag.py
```diff
@@ -130,6 +163,22 @@ class RSPImageTag:
...
+
+    age: datetime.timedelta | None = None
+    """This is a calculated field, rather than a stored one.  This tells us
+    our best guess as to how old an image is and relies on the date field,
+    which may be unknown or inaccurate.
+    """
```
rra, line 180: Since this is a calculated field, it should be a property function or a method, not a data element. That will drastically simplify the code since then you don't have to calculate it in the constructor and you can drop `age` from most of the tests except for an explicit test of age. You can also get rid of the `recalculate_age` method.

A property is simplest, but it has the problem that the basis for calculating the age will keep changing since presumably it would use `datetime.now`. For menu purposes, I don't think you'll want that; you'll want to use the same basis for the whole menu calculation. So I would make this a method that takes a `datetime` argument to serve as the basis point for the age (which could be `None` to use now).

controller/tests/models/domain/rsptag_test.py
```diff
@@ -359,3 +415,11 @@ def test_from_str() -> None:
     # them in dictionary form.
     for tag, expected in test_cases.items():
         assert asdict(RSPImageTag.from_str(tag)) == expected
+
+
+def test_age() -> None:
+    rsptag = RSPImageTag.from_str("d_2021_05_27_20210527")
```
rra, line 421: What's the extra stuff after the normal daily tag pattern?

controller/tests/models/domain/rsptag_test.py
```diff
@@ -359,3 +415,11 @@ def test_from_str() -> None:
...
+    assert rsptag.age is not None
```
rra, line 422: Should check this for accuracy, not just existence.

rra (changes requested):

controller/src/controller/models/domain/rsptag.py
athornton: Almost.  The above is Thursday (which is the day we build).  But yes, fromisocalendar() is what I wanted (and forgot existed)

controller/tests/models/domain/rsptag_test.py
athornton: The tag I copied.  That'd be the "supplementary" part of the tag.  I presume we tossed those in to some of the tags for display name test purposes.  Dropping it will make the intent more clear.

controller/src/controller/models/domain/rsptag.py
athornton: Although, irritatingly, that returns a naive object.

changelog.d/20250203_194233_athornton_DM_48682_dates.md
rra, line 1: These changes aren't user-visible yet, so no change log for these. When this changes something a user would be able to see, that's the PR that should include the change log mesasge.

controller/src/controller/models/domain/rsptag.py
```diff
@@ -157,6 +169,7 @@ def alias(cls, tag: str) -> Self:
             version=None,
             cycle=cycle,
             display_name=display_name,
+            date=None,
```
rra, line 172: Oh, this is a helpful reminder: `RSPImage.resolve_alias` should also set the `date` field on the alias tag if it resolves to a tag with a `date` field that is not `None`, similar to how we override the cycle when we resolve an alias tag. (And that should have a test as well.)

controller/src/controller/models/domain/rsptag.py
```diff
@@ -305,13 +321,16 @@ def _from_match(
...
+        image_date = cls._calculate_date(data)
+
         # Return the results.
         return cls(
             image_type=image_type,
             version=version,
             tag=tag,
             cycle=int(cycle) if cycle else None,
             display_name=display_name,
+            date=image_date,
```
rra, line 333:
```suggestion
            date=cls._calculate_date(data)
```
and remove the `image_date` variable above.

controller/src/controller/models/domain/rsptag.py
```diff
@@ -1,6 +1,7 @@
 """Abstract data types for handling RSP image tags."""
 
 import contextlib
+import datetime
```
rra, line 4:
```suggestion
from datetime import UTC, datetime
```
This is what we do most other places in our code bases.

controller/src/controller/models/domain/rsptag.py
```diff
@@ -349,6 +368,55 @@ def _determine_build(
...
+        if not year:
+            return None
+        week = tagdata.get("week")
+        if year and week:
+            thursday = 4  # We build on Thursday, and Monday is index 1.
+            stamp = datetime.datetime.fromisocalendar(
+                int(year), int(week), thursday
+            )
+            # fromisocalendar() gives us a naive object; force it to UTC
+            return stamp.replace(tzinfo=datetime.UTC)
```
rra, line 395:
```suggestion
            # Weekly images are built on Thursday (ISO weekday 4).
            stamp = datetime.fromisocalendar(int(year), int(week), 4)
            return stamp.replace(tzinfo=UTC)
```
I don't think commenting the `replace` call is needed since its existence communicates the same information as the comment.

controller/src/controller/models/domain/rsptag.py
```diff
@@ -349,6 +368,55 @@ def _determine_build(
...
+        month = tagdata.get("month")
+        day = tagdata.get("day")
+        if not (month and day):
+            return None
+        return datetime.datetime(
+            int(year), int(month), int(day), tzinfo=datetime.UTC
+        )
```
rra, line 402:
```suggestion
        return datetime(int(year), int(month), int(day), tzinfo=UTC)
```

controller/src/controller/models/domain/rsptag.py
```diff
@@ -349,6 +368,55 @@ def _determine_build(
...
+
+        Parameters
+        ----------
+        since
+            Datestamp to measure age from.  If unspecified, the present.
+        """
+        if self.date is None:
+            return None
+        if since is None:
+            since = datetime.datetime.now(datetime.UTC)
```
rra, line 417:
```suggestion
            since = datetime.now(tz=UTC)
```
I guess the `tz=` technically isn't needed but that's what Ruff rewrites things to, so it's what we've been doing most other places.

controller/tests/models/domain/rsptag_test.py
```diff
@@ -168,34 +172,41 @@ def test_from_str() -> None:
...
         "r22_0_0_rc1": {
             "tag": "r22_0_0_rc1",
             "image_type": RSPImageType.CANDIDATE,
             "display_name": "Release Candidate r22.0.0-rc1",
             "version": VersionInfo(22, 0, 0, "rc1"),
             "cycle": None,
+            "date": None,
         },
+        # fromisocalendar() gives you a naive object, even if you
+        # replace the tzinfo.
```
rra, line 186: I don't think this can be correct, and it doesn't seem to match the test.

controller/tests/models/domain/rsptag_test.py
```diff
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+import datetime
```
rra, line 5: Here too.
```suggestion
from datetime import UTC, datetime
```
and adjust the calls accordingly.

controller/tests/models/domain/rsptag_test.py
```diff
@@ -168,34 +172,41 @@ def test_from_str() -> None:
...
         },
+        # fromisocalendar() gives you a naive object, even if you
+        # replace the tzinfo.
         "w_2021_22": {
             "tag": "w_2021_22",
             "image_type": RSPImageType.WEEKLY,
             "display_name": "Weekly 2021_22",
             "version": VersionInfo(2021, 22, 0),
             "cycle": None,
+            "date": datetime.datetime(2021, 6, 3, 0, 0, tzinfo=datetime.UTC),
```
rra, line 193: Here and elsewhere, you omit the hour and minute when creating the date, so I would omit them here as well when comparing the date, since you're relying on the default values of zero. (And you're omitting seconds anyway so may as well omit hours and minutes.)

controller/tests/models/domain/rsptag_test.py
```diff
@@ -359,3 +390,16 @@ def test_from_str() -> None:
...
+    rsptag = RSPImageTag.from_str("d_2021_05_27")
+    assert rsptag.date == datetime.datetime(
+        year=2021, month=5, day=27, tzinfo=datetime.UTC
+    )
+    orig_age = rsptag.age()
+    assert orig_age is not None
+    time.sleep(0.1)
+    new_age = rsptag.age()
+    assert new_age is not None
+    assert new_age > orig_age
```
rra, line 405: I don't think there's any reason to test that the date is changing, and that requires adding delays to the test suite. The important part is that the time depends on the basis passed in and that the math is correct. So, I think the right test here is to pass in a fixed datetime, assert the resulting age is the correct and expected timedelta, grab `datetime.now(tz=UTC)`, call `age()` with no arguments, and assert that the result is within a second of the timedelta you get when subtracting the earlier result of `datetime.now` from `rsptag.date`.

rra (commented):
Mostly small stuff but the change to alias tags reminds me that we need to set the date on alias tags when resolving aliases in `RSPImageTag`. That can be a separate PR if you want.

controller/src/controller/models/domain/rsptag.py
athornton: While I agree, it also doesn't matter because alias tags are not subject to filtering; the only reason to create aliases is in order to pin things to the menu, after all.

controller/tests/models/domain/rsptag_test.py
```diff
@@ -359,3 +387,16 @@ def test_from_str() -> None:
...
+    """Test that the age() method works with and without supplied basis."""
+    rsptag = RSPImageTag.from_str("d_2021_05_27")
+    now = datetime.now(tz=UTC)
+    # If the next three lines take longer than one second, you have a problem.
+    orig_age = rsptag.age(now)
+    assert orig_age is not None
+    new_age = rsptag.age()
+    assert new_age is not None
+    assert (new_age - orig_age) > timedelta(seconds=0)
+    assert (new_age - orig_age) < timedelta(seconds=1)
```
rra, line 402: This still isn't testing the semantics of the function, though. If the implementation of `age` was just `return datetime.now(tz=UTC) - datetime.now(tz=UTC)`, it might pass this test.

```python
    assert rsptag.age(datetime(2021, 5, 28, tzinfo=UTC)) == timedelta(days=1)
    now = datetime.now(tz=UTC)
    expected = now - datetime(2021, 5, 27, tzinfo=UTC)
    assert expected <= rsptag.age() <= expected + timedelta(seconds=1)
```
would test the function itself.

rra (approved):
Mostly looks good, just one remaining comment on the test of `age`. Thanks for fixing the `RSPImage` class!

PR #475, opened 2025-02-06
Add configuration for dropdown filtering

athornton:


controller/src/controller/models/domain/imagepolicy.py
```diff
@@ -0,0 +1,150 @@
...
+
+    'age' means: display any items of the class whose age is the
+    specified age or less.  This age is a duration, specified by a
+    string as accepted by Safir's HumanTimeDelta.  The empty string
+    means "do not filter this class at all."
+
+    'cutoff_version' means: display any items of this class whose semantic
+    version is this large or larger.  This only makes sense for dailies,
+    weeklies, releases, release candidates, and experimentals derived from
+    one of those builds.
```
rra, line 37: Should move to the description of `cutoff_version`. I think it would be shorter to just say that this doesn't apply to unknown tags or experimental tags that aren't derived from a known tag.

controller/src/controller/models/domain/imagepolicy.py
```diff
@@ -0,0 +1,150 @@
...
+
+    'filter_empty' governs the behavior when a policy is applied to an image
+    that does not have data for that category (e.g. an 'age' policy exists,
+    but the image age is 'None' because it could not be determined).  If
+    'filter_empty' is 'True', the image will not be displayed, and if 'False'
+    the image will be displayed.
```
rra, line 43: I would just pick a behavior and delete the option until we have a use case for being able to configure this. `False` seems like a reasonable approach; let's see if it causes any problems.

controller/src/controller/config.py
```diff
@@ -1206,13 +1286,26 @@ class Config(BaseSettings):
...
             description=(
-                "Configuration for which images to prepull and which images to"
+                "Configuration for which images to prepull and to"
                 " display in the spawner menu for users to choose from when"
                 " spawning labs"
             ),
         ),
     ]
 
+    dropdown_menu: Annotated[
```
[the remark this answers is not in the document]
athornton: `mainmenu` and `dropdown` maybe?  Although just gluing in `dropdown_menu` actually corresponds to the structure of the spawner page better.  Maybe the `images` section should be renamed `spawner_page`, since that's what it's really controlling?

Uh, I see the point though.  This all feels kinda clunky.  Let's leave it alone until someone has an idea that feels actually elegant.

controller/src/controller/models/domain/imagepolicy.py
```diff
@@ -0,0 +1,150 @@
...
+from pydantic import BaseModel, BeforeValidator, ConfigDict, Field
+from semver.version import VersionInfo
+
+__all__ = ["IndividualImageClassPolicy", "RSPImagePolicy"]
+
+
+def _empty_str_is_none(inp: Any) -> Any:
+    if isinstance(inp, str) and inp == "":
+        return None
+    return inp
```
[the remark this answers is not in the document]
athornton: Artifact from the reaper CLI, I think?  Anyway, yeah, don't need it.

athornton:
I'm not sure about the last commit.  I was bothered by needing the allow-arbitrary-types thing in the cutoff version field but it feels like I just did a lot of ceremonial boilerplate to make it go away.  Also I do not understand why, if semver publishes basically exactly the boilerplate you need to make Pydantic happy, they do not include that magic by default in semver, or a PydanticVersion class, in the first place.

athornton:
Eh, crap, we do need the policy to be in a separate model because otherwise we get into a circular import problem around RSPTag.

controller/src/controller/config.py
```diff
@@ -86,6 +93,54 @@ def _reject_reserved_paths(v: str) -> str:
     return v
 
 
+class _VersionPydanticAnnotation:
```
rra, line 96: This seems excessive just to be able to store the parsed version number, which I don't think has a lot of benefit. If you keep the type as `str` you can drop all of this. If you want to validate the version, just add a functional validator that checks that you can parse the string with semver.

controller/src/controller/config.py
```diff
@@ -527,6 +582,120 @@ class PrepullerConfig(PrepullerOptions):
     source: DockerSourceConfig | GARSourceConfig
 
 
+class IndividualImageClassPolicy(BaseModel):
```
rra, line 585: I don't think the `Individual` part of this class name is doing much for clarity other than making the class name longer. Maybe just `ImageFilterPolicy` instead? I think that's what this is.

controller/src/controller/config.py
```diff
@@ -527,6 +582,120 @@ class PrepullerConfig(PrepullerOptions):
...
+    """Policy for images to display within a given class.
+
+    The policy has a 'number', an 'age', and a 'cutoff_version' field.  All
+    are optional.
```
rra, line 589: This is obvious from the code and from any documentation Sphinx will generate, so you don't have to write it in the docstring.

rra (changes requested):
There are a bunch of extraneous changes in here that I think are due to branch problems. Rebasing against main will hopefully fix that.

rra:
Yeah, if you want to be able to pass the filter policy in as an argument to methods on `RSPImageTagCollection`, then you'll need to move the filter policy (and just the filter policy) into a separate domain model module. But you still shouldn't need two separate classes; you can just move it out of config into a separate module.

athornton:
> Yeah, if you want to be able to pass the filter policy in as an argument to methods on `RSPImageTagCollection`, then you'll need to move the filter policy (and just the filter policy) into a separate domain model module. But you still shouldn't need two separate classes; you can just move it out of config into a separate module.

I think I disagree.  I want to pass the whole policy (that is, the thing with "release", "weekly", and so on attributes) into the filter method on RSPImageTagCollection and then iterate over the attributes based on RSPImageType.  So the higher-level policy is going to have to live outside config as well as the policy applied to each individual image category (I've decided I like the word "category" better than "class" here as it's slightly less overloaded.

controller/src/controller/config.py
athornton: Yeah, while I was working on this over the weekend (not having seen the review yet) I came to basically that conclusion and just added a string-to-Version validator.

controller/src/controller/config.py
athornton: This one is ImageFilterPolicy and the enclosing one is RSPImageFilterPolicy.  I think that's pretty clear.  I also added an alias attribute to RSPImageFilterPolicy, and the reason why is so that I don't have to special case aliases on the implementation side.

athornton:
> I'm not sure about the last commit. I was bothered by needing the allow-arbitrary-types thing in the cutoff version field but it feels like I just did a lot of ceremonial boilerplate to make it go away. Also I do not understand why, if semver publishes basically exactly the boilerplate you need to make Pydantic happy, they do not include that magic by default in semver, or a PydanticVersion class, in the first place.

Yeah all I really wanted was a string validator.

controller/src/controller/models/domain/imagefilterpolicy.py
```diff
@@ -0,0 +1,133 @@
...
+            description="Policy for daily builds to display.",
+            default_factory=ImageFilterPolicy,
+        ),
+    ]
+
+    release_candidate: Annotated[
+        ImageFilterPolicy,
+        Field(
+            title="Release Candidate",
+            description=("Policy for release candidate builds to display.",),
```
rra, line 110: This is accidentally a tuple.

controller/src/controller/models/domain/imagefilterpolicy.py
```diff
@@ -0,0 +1,133 @@
...
+    )
+
+    alias: Annotated[
+        ImageFilterPolicy,
+        Field(
+            title="Release",
+            description="Policy for releases to display.",
+            default_factory=ImageFilterPolicy,
+        ),
+    ]
```
rra, line 77: I don't think filtering aliases this way really makes sense. I think we want to include the alias iff the underlying tag is included.

rra (approved):
I would drop the alias policy and otherwise just one nit.

rra:
> I think I disagree. I want to pass the whole policy (that is, the thing with "release", "weekly", and so on attributes) into the filter method on RSPImageTagCollection and then iterate over the attributes based on RSPImageType. So the higher-level policy is going to have to live outside config as well as the policy applied to each individual image category (I've decided I like the word "category" better than "class" here as it's slightly less overloaded.)

Oh, yes, you're right, moving the whole thing makes sense.

controller/src/controller/models/domain/imagefilterpolicy.py
athornton: While I agree with you that filtering aliases doesn't make much sense, if we don't special-case the alias category, the implementation of the filter becomes (to my mind) quite a bit cleaner.

https://github.com/lsst-sqre/nublado/blob/05132eff2fb7901b39609256e68340c9702e1b33/controller/src/controller/models/domain/rsptag.py#L636-L655

athornton: If we do special case that, there's an extra test that feels awkward inside the loop.  If you care a lot I'll put it back that way, but that's how I wrote it the first time and it didn't feel as nice as just letting the alias category be handled like all the others.

rra: If you'd rather not implement that filtering for now, I think that's fine, but in that case I think we should just not filter alias tags at all. That avoids the problem of what the right semantic solution is, and we can always add filtering for alias tags later.

PR #481, opened 2025-02-11
tickets/DM-48682-dropdown-implementation: add dropdown menu restriction policy

athornton:
This allows users to restrict the dropdown menu based on per-category policy limiting age, version, or image count.

### DIFF
--- a/controller/src/controller/models/domain/rspimage.py
+++ b/controller/src/controller/models/domain/rspimage.py
@@ -556,8 +556,8 @@ class RSPImageCollection:
 
     def filter(
         self, policy: RSPImageFilterPolicy, age_basis: datetime
-    ) -> Self:
-        """Apply a filter policy and return the remaining tags.
+    ) -> list[RSPImage]:
+        """Apply a filter policy and return the remaining images.
 
         Parameters
         ----------
@@ -568,7 +568,7 @@ class RSPImageCollection:
 
         Returns
         -------
-        RSPImageTagCollection
+        list[RSPImage]
             Tags remaining after policy application.
         """
         images: list[RSPImage] = []
@@ -576,7 +576,7 @@ class RSPImageCollection:
             images.extend(
                 self._apply_category_policy(policy, category, age_basis)
             )
-        return type(self)(images)
+        return images
 
     def _apply_category_policy(
         self,
@@ -588,24 +588,23 @@ class RSPImageCollection:
         remainder: list[RSPImage] = []
         cat_policy = policy.policy_for_category(category)
         if cat_policy is None:
-            remainder.extend(candidates)
-            return remainder
+            return candidates
+        cutoff_date = None
+        if cat_policy.age is not None:
+            cutoff_date = age_basis - cat_policy.age
+        cutoff_version: semver.Version | None = None
+        if cat_policy.cutoff_version is not None:
+            cutoff_version = semver.Version.parse(cat_policy.cutoff_version)
         for image in candidates:
             if cat_policy.number is not None and cat_policy.number <= len(
                 remainder
             ):
                 break
-            if image.date is not None and cat_policy.age is not None:
-                cutoff_date = age_basis - cat_policy.age
+            if image.date is not None and cutoff_date is not None:
                 if image.date < cutoff_date:
                     continue
-            if (
-                image.version is not None
-                and cat_policy.cutoff_version is not None
-            ):
-                img_ver = semver.Version.parse(image.version)
-                cat_pol_ver = semver.Version.parse(cat_policy.cutoff_version)
-                if img_ver < cat_pol_ver:
+            if image.version is not None and cutoff_version is not None:
+                if image.version < cutoff_version:
                     continue
             remainder.append(image)
         return remainder
--- a/controller/src/controller/models/domain/rspimage.py
+++ b/controller/src/controller/models/domain/rspimage.py
@@ -86,6 +86,25 @@ class RSPImage(RSPImageTag):
             **asdict(tag),
         )
 
+    def to_rsptag(self) -> RSPImageTag:
+        """Recover the RSP tag from which this image was derived.
+
+        This is useful for menu calculations.
+
+        Returns
+        -------
+        RSPImageTag
+            RSPImageTag corresponding to this image.
+        """
+        return RSPImageTag(
+            tag=self.tag,
+            image_type=self.image_type,
+            version=self.version,
+            cycle=self.cycle,
+            display_name=self.display_name,
+            date=self.date,
+        )
+
     @property
     def is_possible_alias(self) -> bool:
         """Whether this tag could be an alias."""
--- a/controller/src/controller/models/domain/rspimage.py
+++ b/controller/src/controller/models/domain/rspimage.py
@@ -5,8 +5,12 @@ from __future__ import annotations
 from collections import defaultdict
 from collections.abc import Iterable, Iterator
 from dataclasses import asdict, dataclass, field
+from datetime import datetime
 from typing import Self
 
+import semver
+
+from .imagefilterpolicy import RSPImageFilterPolicy
 from .rsptag import RSPImageTag, RSPImageType
 
 __all__ = [
@@ -84,25 +88,6 @@ class RSPImage(RSPImageTag):
             repository=repository,
             digest=digest,
             **asdict(tag),
-        )
-
-    def to_rsptag(self) -> RSPImageTag:
-        """Recover the RSP tag from which this image was derived.
-
-        This is useful for menu calculations.
-
-        Returns
-        -------
-        RSPImageTag
-            RSPImageTag corresponding to this image.
-        """
-        return RSPImageTag(
-            tag=self.tag,
-            image_type=self.image_type,
-            version=self.version,
-            cycle=self.cycle,
-            display_name=self.display_name,
-            date=self.date,
         )
 
     @property
@@ -568,3 +553,59 @@ class RSPImageCollection:
                     other.aliases.add(new.tag)
             old.aliases.add(new.tag)
             self._by_digest[new.digest] = new
+
+    def filter(
+        self, policy: RSPImageFilterPolicy, age_basis: datetime
+    ) -> Self:
+        """Apply a filter policy and return the remaining tags.
+
+        Parameters
+        ----------
+        policy
+            Policy governing tag filtering.
+        age_basis
+            Timestamp to use as basis for image age calculation.
+
+        Returns
+        -------
+        RSPImageTagCollection
+            Tags remaining after policy application.
+        """
+        images: list[RSPImage] = []
+        for category in RSPImageType:
+            images.extend(
+                self._apply_category_policy(policy, category, age_basis)
+            )
+        return type(self)(images)
+
+    def _apply_category_policy(
+        self,
+        policy: RSPImageFilterPolicy,
+        category: RSPImageType,
+        age_basis: datetime,
+    ) -> list[RSPImage]:
+        candidates = list(self._by_type[category])
+        remainder: list[RSPImage] = []
+        cat_policy = policy.policy_for_category(category)
+        if cat_policy is None:
+            remainder.extend(candidates)
+            return remainder
+        for image in candidates:
+            if cat_policy.number is not None and cat_policy.number <= len(
+                remainder
+            ):
+                break
+            if image.date is not None and cat_policy.age is not None:
+                cutoff_date = age_basis - cat_policy.age
+                if image.date < cutoff_date:
+                    continue
+            if (
+                image.version is not None
+                and cat_policy.cutoff_version is not None
+            ):
+                img_ver = semver.Version.parse(image.version)
+                cat_pol_ver = semver.Version.parse(cat_policy.cutoff_version)
+                if img_ver < cat_pol_ver:
+                    continue
+            remainder.append(image)
+        return remainder
--- a/controller/src/controller/models/domain/rsptag.py
+++ b/controller/src/controller/models/domain/rsptag.py
@@ -107,8 +107,8 @@ class RSPImageTag:
     image_type: RSPImageType
     """Type (release series) of image identified by this tag."""
 
-    version: str | None
-    """Version information as a string representing a semantic version."""
+    version: semver.Version | None
+    """Version information as a semantic version."""
 
     cycle: int | None
     """XML schema version implemented by this image (only for T&S builds)."""
@@ -298,14 +298,12 @@ class RSPImageTag:
 
         # Construct the semantic version.  It should be impossible, given our
         # regexes, for this to fail, but if it does that's handled in from_str.
-        version = str(
-            semver.Version(
-                major=int(major),
-                minor=int(minor),
-                patch=int(patch),
-                prerelease=pre,
-                build=build,
-            )
+        version = semver.Version(
+            major=int(major),
+            minor=int(minor),
+            patch=int(patch),
+            prerelease=pre,
+            build=build,
         )
 
         # If there is extra information, add it to the end of the display name.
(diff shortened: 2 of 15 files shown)


athornton:
Ugggh, gonna have to figure out how to do the JSONSchema bit for Semver.

controller/src/controller/models/domain/rsptag.py
```diff
@@ -576,3 +578,86 @@ def subset(
...
+            if tag_age is not None and cat_policy.age is not None:
+                if tag_age > cat_policy.age:
+                    continue
+            # Apply version policy
+            if (
+                tag.version is not None
+                and cat_policy.cutoff_version is not None
+            ):
+                if tag.version < cat_policy.cutoff_version:
+                    continue
```
rra, line 660: Looks like you don't need the `age` method on an `RSPImageTag` at all (or the `by_type` method above). This method does need to take an argument giving the basis for the age calculation so that it's consistent for all calculations. I'm calling that `age_basis` here.

```suggestion
        cutoff_version = policy.cutoff_version
        cutoff_date = None
        if policy.age:
            cutoff_date = age_basis - policy.age
        for tag in self._by_type[category]:
            if policy.number is not None and len(remainder) == policy.number:
                break
            if cutoff_version is not None:
                if tag.version is not None and tag.version < cutoff_version:
                    continue
            if cutoff_date and tag.date and tag.date < cutoff_date:
                continue
```

This will get more of the code on one line and should make it easier to follow. The comments are just repeating the code, so I'm not sure they add much.

controller/src/controller/models/domain/rsptag.py
```diff
@@ -576,3 +578,86 @@ def subset(
...
+            RSPImageType to select from collection.
+
+        Returns
+        -------
+        RSPImageCollection
+            Collection of tags matching the given type.
+        """
+        return type(self)(self._by_type[image_type])
+
+    def apply_policy(self, policy: RSPImageFilterPolicy) -> Self:
```
rra, line 597: I would call this method `filter`. That's exactly what it does; this seems a bit wordy.

controller/src/controller/models/domain/rsptag.py
```diff
@@ -576,3 +578,86 @@ def subset(
...
+        """Apply a filter policy and return the remaining tags.
+
+        Note that this relies on the ordering of types, as subset() does.
```
rra, line 600: This is an invariant maintained by the class and the caller doesn't need to know about it, so it shouldn't be in the docstring.

controller/src/controller/services/source/gar.py
```diff
@@ -144,11 +149,20 @@ def menu_images(self) -> list[MenuImage]:
...
+            All known images meeting filter criteria.
         """
+        # This is kind of clunky.  I feel like the implementation of the
+        # storage layer threw away the tags too soon.  That said, adding a
+        # to_rsptag() method to MenuImage lets us do this in a straightforward
+        # if not very elegant way.
+        all_tags = [x.to_rsptag() for x in list(self._images.all_images())]
+        all_coll = RSPImageTagCollection(all_tags)
+        filtered_coll = all_coll.apply_policy(self._image_filter)
+        filtered_tags = list(filtered_coll.all_tags())
```
rra, line 161: I think this is too clunky. It's a lot cleaner semantically to add a `filter` method to `RSPImageCollection` instead. That means either a bit of code duplication or moving filtering of each class of images into a generic helper function, but it means we have a clean abstract data type and aren't doing complicated workarounds. As above, this should pass in `datetime.now(tz=UTC)` as the basis for tag ages.

controller/tests/data/standard/output/lab-form.html
```diff
@@ -34,8 +34,6 @@
       <option value="lighthouse.ceres/library/sketchbook:recommended@sha256:5678">Recommended (Weekly 2077_43)</option>
       <option value="lighthouse.ceres/library/sketchbook:w_2077_43@sha256:5678">Weekly 2077_43</option>
       <option value="lighthouse.ceres/library/sketchbook:d_2077_10_23@sha256:1234">Daily 2077_10_23</option>
-      <option value="lighthouse.ceres/library/sketchbook:latest_weekly">latest_weekly</option>
-      <option value="lighthouse.ceres/library/sketchbook:latest_daily">latest_daily</option>
```
rra, line 38: Oh, this is the reason why a policy of not including unknown tags doesn't work. If we use a Docker source, all alias tags are unknown tags unless they correspond to a pre-pulled image. I'm not sure how I feel about that. I think it means we need to include all unknown tags for as long as we support Docker sources.

controller/tests/models/domain/imagefilter_test.py
```diff
@@ -0,0 +1,191 @@
...
+def test_imagefilter() -> None:
+    """Test behavior of an RSPImageTagCollection object under a filter."""
+
+    # Make some weekly and daily tags.  Because we can run this test at
+    # an arbitrary time...we need to dynamically generate these.
+    def to_weekly_tag(dt: datetime) -> str:
+        ic = dt.isocalendar()
+        icy = str(ic.year).zfill(2)
+        icw = str(ic.week).zfill(2)
+        return f"w_{icy}_{icw}"
```
rra, line 29:
```suggestion
    def to_weekly_tag(time: datetime) -> str:
        year, week, _ = time.isocalendar()
        return f"w_{year:d}_{week:02d}"
```

controller/tests/models/domain/imagefilter_test.py
```diff
@@ -0,0 +1,191 @@
...
+
+    def to_daily_tag(dt: datetime) -> str:
+        mo = str(dt.month).zfill(2)
+        dy = str(dt.day).zfill(2)
+        return f"d_{dt.year}_{mo}_{dy}"
```
rra, line 34:
```suggestion
    def to_daily_tag(time: datetime) -> str:
        return f"d_{time.year:d}_{time.month:02d}_{time.day:02d}"
```

controller/tests/models/domain/imagefilter_test.py
```diff
@@ -0,0 +1,191 @@
...
+
+    # Because image ages are calculated based on the tags, and are
+    # therefore calculated as if they were born at midnight UTC, you
+    # might have one fewer image than you expect in a category depending
+    # on local time if you're west of UTC, or one more if you're east of UTC.
```
rra, line 39: If that's the case there's some serious flaw in the code somewhere because nothing in either the controller or in the tests should be based on local time zones. Every operation should be done in UTC from beginning to end, so nothing related to time zones should matter.

There therefore shouldn't be any ranges; we should be able to determine the expected output exactly.

controller/tests/models/domain/imagefilter_test.py
```diff
@@ -0,0 +1,191 @@
...
+            # dated releases we use calver
+            cutoff_version=Version(
+                major=last_week.isocalendar().year,
+                minor=last_week.isocalendar().week,
+                patch=0,
+            )
+        ),
+    )
+
+    filtered = collection.apply_policy(policy)
```
rra, line 138: `apply_policy` (or `filter` per above) should take a time basis for the filtering, so you can pass in `now`. That will fix your problems with having a variable number of tags in some category and simplify a bunch of the rest of the test suite, since then you can probably just check the result of `all_tags` against an expected list instead of checking each class separately.

controller/src/controller/services/source/docker.py
```diff
@@ -201,12 +205,13 @@ def menu_images(self) -> list[MenuImage]:
...
         -------
         list of MenuImage
-            All known images.
+            All known images meeting filter criteria.
         """
         registry = self._config.registry
         repository = self._config.repository
         menu_images = []
-        for tag in self._tags.all_tags():
+        filtered_tags = self._tags.apply_policy(self._image_filter)
```
rra, line 213: Should pass in `datetime.now(tz=UTC)` as the basis for tag ages.

rra (changes requested):

controller/tests/models/domain/imagefilter_test.py
athornton: I think the problem is that now() might be after midnight the next day, or before midnight of the day the image was presumably minted, depending on where you actually are.  That means an image can be a negative duration old, and "yesterday" and "today" don't mean (with respect to the image age) what they actually should.

I mean, I think maybe where I'm going with this is that maybe "age", since we can't really get the image age, the best we can do is "midnight UTC of the date in the tag" or perhaps "upload date" or "distrib tag creation date", is not a good filter choice.  Maybe we really can do everything we need to do with image count and semver.

controller/tests/data/standard/output/lab-form.html
rra: latest_weekly and latest_daily are currently alias tags that only appear in the dropdown and are not displayed on the main menu. Currently the underlying images themselves are always prepulled given our current prepull policy, so even with Docker the tags would be resolved to alias tags, but this would not be the case if, say, the menu was configured to hide dailies. But I think in that case we wouldn't want to make latest_daily disappear from the dropdown, would we?

athornton: ....I...don't know?  I guess I'll make unknown, like alias, return `None` and therefore show all the unknown tags, but yuck, and that means I need to clean up Ross's test tag in GAR.

I really don't like showing all of the broken tags we will have accumulated over the years just to be able to show latest_daily/latest_weekly in the (presumably rare) cases we don't prepull them.

controller/src/controller/services/source/docker.py
athornton: I think I prefer leaving this alone and defaulting the time basis in the filter() method to `now()` but I'm certainly willing to be told why I'm wrong.

controller/tests/models/domain/imagefilter_test.py
athornton: Oh, nm, now I get why I shouldn't default it in `filter()`

controller/src/controller/services/source/docker.py
athornton: Oh, you said that a couple comments up.

controller/src/controller/services/source/gar.py
athornton: I don't understand this comment.  The problem I'm trying to solve here is that the storage layer gives us back a collection that, when iterated, emits `MenuImage`, but the thing we want for the dropdown is a collection of RSPImageTags.  So either I need to rewrite the storage layer to emit Tags, or I need a way to recover a tag from a MenuImage.  Or I need to pack the Tag directly into the MenuImage but then serialization becomes harder and we end up with duplicated fields.

controller/tests/data/standard/output/lab-form.html
athornton: ...I guess we could just put an explicit policy on Unknown again, so I can set it to `number: 0` for GAR...

controller/src/controller/services/source/gar.py
athornton: I feel like there's too much of an impedance mismatch between GAR and Docker drivers here, but I also don't feel like I really want to embark on that refactor yet.

The Docker v2 API does at least exist now, although it is not (I think) fully supported by Docker Hub(!!) and I have no idea whether Nexus et al support it.  But I think maybe we should defer this bit of work until (and unless) we tackle Docker registries again.

controller/src/controller/models/domain/rsptag.py
athornton: Fair, although I prefer doing the cutoff_date and version-to-semver calculations inside the "do we have a policy for this kind of filter, and do we have enough info on the tag to apply it" bit, when we're already committed to applying that filter.

controller/tests/models/domain/imagefilter_test.py
athornton: no longer relevant

controller/src/controller/services/source/gar.py
athornton: Oh, never mind.  Add a `filter()` to `RSPImageCollection` too, that does the right thing, and maybe factor out the actual policy application to an external function.  I understand now.

controller/tests/data/standard/output/lab-form.html
athornton: I'm leaving it at `None` to match existing behavior, but...I'm ag'in' it.

controller/src/controller/models/domain/rspimage.py
```diff
@@ -549,3 +553,59 @@ def _resolve_duplicate(self, new: RSPImage, old: RSPImage) -> None:
...
+        policy: RSPImageFilterPolicy,
+        category: RSPImageType,
+        age_basis: datetime,
+    ) -> list[RSPImage]:
+        candidates = list(self._by_type[category])
+        remainder: list[RSPImage] = []
+        cat_policy = policy.policy_for_category(category)
+        if cat_policy is None:
+            remainder.extend(candidates)
+            return remainder
```
rra, line 592:
```suggestion
        if cat_policy is None:
            return candidates
```

controller/src/controller/models/domain/rspimage.py
```diff
@@ -549,3 +553,59 @@ def _resolve_duplicate(self, new: RSPImage, old: RSPImage) -> None:
...
+                if image.date < cutoff_date:
+                    continue
+            if (
+                image.version is not None
+                and cat_policy.cutoff_version is not None
+            ):
+                img_ver = semver.Version.parse(image.version)
+                cat_pol_ver = semver.Version.parse(cat_policy.cutoff_version)
+                if img_ver < cat_pol_ver:
+                    continue
```
rra, line 609: I'd still like to see the threshold calculations factored out of the loop rather than recalculating the cutoff date on every time through the loop. I don't care about the performance hit, but I think it would make the code easier to read.

controller/src/controller/models/domain/rsptag.py
```diff
@@ -122,8 +107,8 @@ class RSPImageTag:
     image_type: RSPImageType
     """Type (release series) of image identified by this tag."""
 
-    version: Version | None
-    """Version information as a semantic version."""
+    version: str | None
+    """Version information as a string representing a semantic version."""
```
rra, line 111: No, I think this change is wrong. This should stay a `Version`. It shouldn't be necessary to serialize this object. That will remove the need for a bunch of the changes below.

controller/src/controller/models/domain/rsptag.py
```diff
@@ -576,3 +551,59 @@ def subset(
...
+            Tags remaining after policy application.
+        """
+        tags: list[RSPImageTag] = []
+        for category in RSPImageType:
+            tags.extend(
+                self._apply_category_policy(policy, category, age_basis)
+            )
+        return type(self)(tags)
+
+    def _apply_category_policy(
```
rra, line 579: Same comments here as above.

controller/src/controller/models/domain/rspimage.py
```diff
@@ -549,3 +553,59 @@ def _resolve_duplicate(self, new: RSPImage, old: RSPImage) -> None:
                     other.aliases.add(new.tag)
             old.aliases.add(new.tag)
             self._by_digest[new.digest] = new
+
+    def filter(
+        self, policy: RSPImageFilterPolicy, age_basis: datetime
+    ) -> Self:
```
rra, line 559: Given how `filter` is used, I think it should just return `list[RSPImageTag]` or `list[RSPImage]`. None of the callers use the collection properties; they just immediately turn this into a list, so turning it into a collection and then back into the same list is unnecessary work.

rra (changes requested):

controller/src/controller/models/domain/rsptag.py
athornton: I'm pretty sure I got into this mess in the first place by having it be a `Version` internally.  I'll give it a try, but I think it's going to fail.

athornton: Oh, maybe I see, leave it a str in the policy (and do the parse()) but a Version on the image/tag

controller/src/controller/models/domain/rspimage.py
athornton: I don't, but I'm willing to change it.

rra (approved):

PR #487, opened 2025-02-19
DM-49032: 8.3 backport: fix another `websockets` >=14 breaking change

fajpunk:
Version 14 of `websockets` [changed the signature of the `connect` method](https://websockets.readthedocs.io/en/stable/howto/upgrade.html#extra-headers-additional-headers)

Without this, Mobu breaks with:
```
TypeError
create_connection() got an unexpected keyword argument 'extra_headers'
```

With this change, mobu is able to use the nublado client to execute notebooks again (tested locally).


fajpunk:
No need to backport, gonna release off of main.

PR #490, opened 2025-02-25
Bump safir upper bound now that 10.0.0 is released

fajpunk:
The breaking change with Safir `10` is that Python 3.11 is no longer supported.

### DIFF
--- a/changelog.d/20250224_201855_danfuchs.md
+++ b/changelog.d/20250224_201855_danfuchs.md
@@ -1,0 +1,5 @@
+<!-- Delete the sections that don't apply -->
+
+### Other changes
+
+- Nublado client: raise upper bound on safir dependency to `<11`.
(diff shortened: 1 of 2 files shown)


changelog.d/20250224_201855_danfuchs.md
rra, line 1: We usually don't bother with a change log for dependency bound changes in libraries, although I can see the argument either way for this one since it does affect the downstream packages. (And now I'm of two minds about putting a version pin on Safir dependencies in the Nublado client, hm.)

rra (approved):

changelog.d/20250224_201855_danfuchs.md
fajpunk: Yeah, I was thinking it's different than a service dep change, because it could be good to know for consumers of the lib, though I could go either way too. As for pinning Safir, my vote is we keep the pin for now and we can revisit if it gets to be more painful? It definitely is a pain to have to cut a new Nublado client release for every major Safir release, but theoretically, we'd have to verify compatibility anyway.

PR #494, opened 2025-03-10
DM-49388: Support per-user subdomains in the Nublado client

rra:
In the Nublado client, check the first redirect when accessing JupyterHub or JupyterLab and see if it is a redirect to a different host. If so, update the internal base URL of JupyterHub or JupyterLab to the new base URL. This ensures that subsequent calls that cannot be redirected (`POST` or `DELETE`, for instance) will use the correct base URL and not fail due to a redirect or missing cookies.

Reorganize the Nublado client code to group together internal methods. Reduce the debug logging for searching for XSRF cookies to only note when a new XSRF cookie was found, instead of logging every step of the process.

Refactor the Nublado client mock to start pulling common checks into a shared static method used as a decorator. Not all the checks that could be moved have yet been moved; that will happen in a future change with a broader refactoring.


client/src/rubin/nublado/client/testing/_jupyter.py
```diff
@@ -296,41 +304,49 @@ def missing_lab(self, request: Request) -> Response:
...
-                    "Location": _url(
-                        self._base_url, f"user/{user}/oauth_callback"
-                    ),
+                    "Location": self._url(f"user/{user}/oauth_callback"),
                     "Set-Cookie": xsrf,
                 },
             )
         else:
+            if self._use_subdomains:
+                host = "nb." + self._base_url.host
```
athornton, line 344: Is `nb.` actually constant?  I guess effectively it is, whether or not it comes from something settable in values.yaml.

client/src/rubin/nublado/client/testing/_jupyter.py
```diff
@@ -414,9 +446,17 @@ def get_content(self, request: Request) -> Response:
         user = self._get_user_from_headers(request)
         if user is None:
             return Response(403, request=request)
+        if redirect := self._maybe_redirect(request, user):
+            return Response(
+                302, request=request, headers={"Location": redirect}
+            )
```
athornton, line 452: All this repetitive boilerplate is making me nostalgiac for Go.

Yeah, I get that since we have to return from the function if this request generates a redirect, there's no better way to do this.

A way to shorten this and reduce the duplication might be to convert _maybe_redirect into something that raised an Exception that had the redirect as a field, and then put everything here inside a try: if you catch that Exception, return the redirect.  That might be a lot less clear though.

client/src/rubin/nublado/client/testing/_jupyter.py
rra: Yeah, for the test mock it is constant, similar to how the test mock hard-codes `/nb`. I don't think we would gain anything from making that configurable in tests. I might be missing something, though, and if so, let me know!

client/src/rubin/nublado/client/testing/_jupyter.py
```diff
@@ -154,53 +164,89 @@ def register_extension_result(
...
-    def _extract_user_from_mock_token(token: str) -> str:
-        # remove "gt-", and split on the dot that marks the secret
-        return urlsafe_b64decode(token[3:].split(".", 1)[0]).decode()
-
-    def login(self, request: Request) -> Response:
-        user = self._get_user_from_headers(request)
-        if user is None:
-            return Response(403, request=request)
-        if JupyterAction.LOGIN in self._fail.get(user, {}):
+    def _check(
```
athornton, line 177: Huh.  I don't think I would have thought about turning this into a decorator.  Cool.

client/src/rubin/nublado/client/nubladoclient.py
```diff
@@ -1192,37 +1188,65 @@ def _extract_xsrf(self, response: Response) -> str | None:
...
                 status_code=response.status_code,
-                # Logging the set-cookie header can be useful here but it
-                # leaks secrets.  Don't put that code in a release build.
             )
         return xsrf
 
-    def _url_for(self, partial: str) -> str:
-        """Construct a JupyterHub or Jupyter lab URL from a partial URL.
+    def _initialize_client(self) -> None:
+        """Initialize the HTTP client connection pool for JupyterHub.
```
athornton, line 1201: I was a little worried when that long discursive comment about connection pools above was removed, but this is a much clearer formulation of the issue.

inithome/src/rubin/nublado/inithome/provisioner.py
```diff
@@ -76,14 +77,14 @@ async def provision(self) -> None:
...
-            logging.warning(f"{msg} but is empty, resetting ownership")
+            self._logger.warning(f"{msg} but is empty, resetting ownership")
             os.chown(self._home, self._uid, self._gid)
 
         # Check mode. We only care about the permission bits.
         mode = stat.st_mode & 0o777
         if mode != 0o700:
             msg = f"{self._home} has unexpected permissions: 0{mode:o} != 0700"
-            logging.warning(msg)
+            self._logger.warning(msg)
```
athornton, line 87: Ah, I see you updated Ruff too!  I found this one independently yesterday on DM-48780.

athornton (approved):
Looks good.

PR #501, opened 2025-03-21
DM-49624: Add configuration option for JupyterHub logout URL

rra:
In the JupyterHub authenticator, once the user logs out, we want to redirect them to the Gafaelfawr logout handler. This previously was done with a relative redirect to `/logout`. This fails with per-user subdomains, since this redirect may be to the JupyterHub hostname, which will then redirect back to the JupyterHub logout URL and create an infinite redirect loop. Add a configuration option to the authenticator that we can set in Phalanx to the URL of the Gafaelfawr logout route.


athornton (approved):
LGTM

PR #502, opened 2025-03-22
DM-49644: Start the event watcher before object creation

rra:
When creating a user's lab, start the event watcher immediately after creating the namespace and before creating any other objects. This will hopefully allow us to show the user more events as they happen instead of only displaying them in bulk after the pod has been created.


rra:
This was a nice idea, but it turns out to not work. Even if we watch all events in the namespace, the events don't show up until the pod has been created, so there's no change in the timing.

PR #514, opened 2025-03-28
DM-49774: Push inithome containers to Google

rra:
GitHub is rate-limiting fetches of the inithome container during scaling tests. Also push it to Google so that we can hopefully get the benefit of a higher rate limit.


athornton (approved):
LGTM

PR #515, opened 2025-04-07
Always push new tagged JL-base images

athornton:


athornton:
So, I think what has been going on is that when we tag the image, there are no changes to Dockerfile or any of the files in jupyterlab-base, so it doesn't pass the filter stage, so we don't do a build-and-push.

This change (I think) is going to force a build-and-push on a new tag, whether or not we have changed anything.

rra (approved):
At some point we probably should move everything to push on release instead of push on tag, but I think this is correct and consistent with how we're handling the other Docker images. (There's also a better way to do path filtering that allows you to not reiterate the conditional everywhere that I found for Gafaelfawr, but we can fix that later.)

PR #521, opened 2025-04-28
Add 'latest' to jupyterlab-base release tags

athornton:


jonathansick (approved):
Yep, I think this will work.

PR #527, opened 2025-05-21
Force rebuild of jupyterlab-base with newer input containers

athornton:


athornton:
Test image with new RJE ran fine, extensions worked.

PR #531, opened 2025-06-02
Add documentation for RSP-image updates

athornton:
Document how we get new features into the Lab container.


jonathansick (approved):
Thanks for writing this down.

- I think this page should go under "Developers" section of the docs because making new nublado release seems like a developer thing rather than an administrator thing.
- I think the really important sections here are testing and releasing; the lsst-rsp and rsp-jupyter-extensions sections could almost better be part of those packages' READMEs/docs. But the testing and release process is what really requires an understanding of the Nublado repo ecosystem and build process.
- It could be useful to show samples of the `install-rsp-user` where branches of rsp-jupyter-extensions are being installed.

PR #533, opened 2025-06-11
Pass homedir schema to environment

athornton:


rra (changes requested):
I don't think this belongs here; the controller doesn't use it, it's specific to one environment, and adding it here makes it part of our API for lab containers going forward, but it's very specific to a workaround for USDF. Instead, what about just setting it directly in the values file for USDF environments?

If we want to formalize this as a feature, I don't think we should hard-code the understanding that the homedir schema affects scratch. We should instead pass down the path to scratch directly. But the quick workaround is to just do this in the environment values file.

athornton:
Fair enough.  Injecting it directly at usdf would certainly do the trick.

PR #537, opened 2025-06-22
Simplify WebDAV UI text

frossie:
I am not sure I understand the difference between 
`controller/tests/data/fileserver/output/fileserver.html`
and
`controller/tests/data/standard/output/fileserver.txt`
but I am sure I will find out.

I also don't understand why the docs test sometimes fails  with (`raphviz.backend.execute.ExecutableNotFound: failed to execute PosixPath(\'dot\'), make sure the Graphviz executables are on your systems\' PATH\n'`) but then passes with no changes


frossie:
Okay cool, just take over this branch for the scope change so that I can merge it at the same time as the docs update reflecting the change?

rra:
Should now be done along with the change log. I'll make a new release once this is merged.

PR #539, opened 2025-06-24
DM-51529: Add debugging for file server reconciliation

rra:
Log more details when retrieving the list of running file servers, since we seem to be missing file servers we should be finding.


controller/src/controller/services/fileserver.py
```diff
@@ -236,38 +236,103 @@ async def reconcile(self) -> None:
...
             with contextlib.suppress(UnknownUserError):
                 await self.delete(username)
-        for username in invalid:
+
+    async def _delete_unexpected_servers(
+        self, to_delete: set[str], start: datetime
+    ) -> None:
+        """Delete invalid servers that weren't expected to be running.
+
+        This is not necessarily an error case, since it can happen during
```
athornton, line 320: Oh.  I wouldn't have thought of that.

athornton (approved):
This all seems good, and I find the logic easier to follow, largely because of the commentary.

PR #541, opened 2025-06-24
DM-51529: Add config options for refresh intervals

rra:
Move the constants that controlled the refresh interval for image sources and node image caches and the reconcile intervals for file servers and labs into configuration options. This allows reducing the intervals when debugging and more easily overriding the intervals in the test suite.


athornton (approved):
That all looks good.

PR #542, opened 2025-06-25
DM-51529: Fix race condition with file server reconciliation

rra:
If a file server was in the process of being created before background reconciliation started and reconciliation ran while the ingress was invalid, reconcile would delete the file server being spawned. Handle this case by adding more state to track operations in progress and skip file servers with operations in progress while reconciling.

Testing this uncovered an additional problem of a race condition on deleting file servers during reconciliation, fixed by ignoring unknown user exceptions, and an error in the delay logic in background loops that didn't handle very short intervals, such as those now used by the test suite.

### DIFF
--- a/controller/requirements/dev.in
+++ b/controller/requirements/dev.in
@@ -26,6 +26,6 @@ types-pyyaml
 
 # Documentation
 autodoc_pydantic
-documenteer[guide]>=1
+documenteer[guide]>=1.4.5
 scriv
 sphinx-diagrams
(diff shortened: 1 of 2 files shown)


controller/src/controller/services/fileserver.py
```diff
@@ -228,16 +254,16 @@ async def reconcile(self) -> None:
...
+                if self._servers[username].modified_since(start):
                     continue
             if username in to_delete:
                 msg = "Removing broken fileserver for user"
                 self._logger.warning(msg, user=username)
             else:
                 msg = "No file server job for user, removing remnants"
                 self._logger.warning(msg, user=username)
-            await self.delete(username)
+            with contextlib.suppress(UnknownUserError):
```
athornton, line 265: I don't think I understand in what case delete(username) could throw the UnknownUserError.  If there's some way multiple things can be working on the user at once, do we want some explicit locking mechanism?

athornton (approved):
I have a couple questions, but I'm pretty sure they are due to my lack of understanding of what's going on in the control flow, rather than anything wrong with the code.

PR #543, opened 2025-06-25
DM-51556: Allow configuring the default lab size

rra:
Add a new `controller.config.lab.defaultSize` configuration option that sets the default lab size explicitly. If it is not set, or is set to a size the user's quota does not allow them to spawn, the default remains the first-listed size that the user's quota allows them to spawn.

### DIFF
--- a/controller/src/controller/handlers/form.py
+++ b/controller/src/controller/handlers/form.py
@@ -56,6 +56,16 @@ async def get_user_lab_form(
     else:
         sizes = config.lab.sizes
 
+    # Determine the default size.
+    default_size = None
+    if config.lab.default_size:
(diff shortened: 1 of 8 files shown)


controller/src/controller/handlers/form.py
```diff
@@ -56,6 +56,16 @@ async def get_user_lab_form(
     else:
         sizes = config.lab.sizes
 
+    # Determine the default size.
```
athornton, line 59: I like this.

athornton (approved):
This looks great.  I particularly appreciate the validator so things keep working even if we change the sizes and forget to change the default.

PR #549, opened 2025-07-07
tickets/DM-51698: Add limit_to_request_ratio config item

athornton:
Make limit_to_request_ratio configurable rather than constant.


fajpunk:
As long as we're changing things, what do you think about ditching the ratio and allowing both limits and requests to be directly specified?

athornton:
Not a fan, because then you have to do it once per size option.

PR #551, opened 2025-07-23
tickets/DM-51906: Add RSP version information to RSPTag

athornton:


rra (commented):
Unfortunately, I don't think this approach of overloading the pre-release field can work. We want the sort order:
```
r21_1_1 < r_21_1_1_rsp2.1.0 < r21_1_1_rsp10.4.5
```
and I believe currently both of those comparisons fail (return the opposite of what they should).

I also noticed that this adds the RSP version to the weeklies and dailies, but I thought we said we didn't want to do that because we don't know what version of dependencies we will pick up and therefore the rsp1.2.0 of a release build and the rsp1.2.0 of a weekly or daily build will be different things and may not produce the same results, which would be confusing? Although I guess that's also true of *any* two builds since we're not pinning dependencies. I'm not sure what Frossie wants to do about that.

athornton:
Added a test showing that r_21_1_1_rsp2.1.0 < r21_1_1_rsp10.4.5

athornton:
(I am not concerned about tags on weekly and daily versions in that we typically would not bother to rebuild them with an updated RSP version; we also could only inject the tag when we build a release rather than a weekly or daily, which feels to me like a better way to handle it than in the tag class.)

athornton:
OK, I think that will do it.  RSP version is its own 3-component semver now, and it's been inserted into the right place in the sort order.

controller/src/controller/models/domain/rsptag.py
```diff
@@ -49,8 +51,22 @@
...
 _TAG_REGEXES = [
+    # r23_0_0_rc1_rsp1.2.3_c0020.001_20210513
+    (
+        RSPImageType.CANDIDATE,
+        re.compile(_CANDIDATE + _RSP + _CYCLE + _REST + "$"),
+    ),
+    # r23_0_0_rc1_rsp1.2.3_c0020.001
+    (RSPImageType.CANDIDATE, re.compile(_CANDIDATE + _RSP + _CYCLE + "$")),
+    # r23_0_0_rsp1.2.3_rc1_20210513
+    (RSPImageType.CANDIDATE, re.compile(_CANDIDATE + _RSP + _REST + "$")),
```
rra, line 67: Comments for the first two version numbers are wrong, I think (the `rc` part is before the `rsp` part, surely).

controller/src/controller/models/domain/rsptag.py
```diff
@@ -402,10 +511,13 @@ def _compare(self, other: object) -> int:
             0 if equal, -1 if self is less than other, 1 if self is greater
             than other, `NotImplemented` if they're not comparable.
         """
-        if not isinstance(other, RSPImageTag):
-            return NotImplemented
-        if self.image_type != other.image_type:
+        try:
+            self._check_comparable_types(other)
+        except NotImplementedError:
             return NotImplemented
```
rra, line 517: A helper that returns a boolean seems simpler:
```suggestion
        if not self._is_comparable_type(other):
            return NotImplemented
```

controller/src/controller/models/domain/rsptag.py
```diff
@@ -402,10 +511,13 @@ def _compare(self, other: object) -> int:
...
+        # If we got this far, we know they are both RSPImageTags even if
+        # mypy doesn't.
+        other = cast("RSPImageTag", other)
```
rra, line 520: If you're going to move the check to a method, you need to use a type guard so that the type inference works. That will avoid the cast and ensure the typing still works properly. See [the mypy documentation](https://mypy.readthedocs.io/en/stable/type_narrowing.html#user-defined-type-guards), the short version of which is in the suggestion below on the helper function.

controller/src/controller/models/domain/rsptag.py
```diff
@@ -414,19 +526,50 @@ def _compare(self, other: object) -> int:
...
         # since we want newer cycles to sort ahead of older cycles (and newer
         # cycle builds to sort above older cycle builds) in otherwise matching
         # tags, and the cycle information is stored in the build.
+        return self._compare_build_versions(other)
+
+    def _check_comparable_types(self, other: object) -> None:
+        if not isinstance(other, RSPImageTag):
+            raise NotImplementedError
+        if self.image_type != other.image_type:
+            raise NotImplementedError
```
rra, line 548:
```suggestion
    def _is_comparable_type(self, other: object) -> TypeGuard[RSPImageTag]:
        if not isinstance(other, RSPImageTag):
            return False
        return self.image_type == other.image_type:
```

controller/src/controller/models/domain/rsptag.py
```diff
@@ -414,19 +526,50 @@ def _compare(self, other: object) -> int:
...
+
+    def _compare_rsp_versions(self, other: Self) -> int:
+        if self.rsp_version is None or other.rsp_version is None:
+            if self.rsp_version is None:
+                if other.rsp_version is not None:
+                    return -1
+                return 0
+            if other.rsp_version is None:
+                return 1
+        return self.rsp_version.compare(other.rsp_version)
```
rra, line 558:
```suggestion
    def _compare_rsp_versions(self, other: Self) -> int:
        if self.rsp_version is None:
            return 0 if other.rsp_version is None else -1
        elif other.rsp_version is None:
            return 1
        else:
            return self.rsp_version.compare(other.rsp_version)
```

controller/src/controller/models/domain/rsptag.py
```diff
@@ -414,19 +526,50 @@ def _compare(self, other: object) -> int:
...
+
+    def _compare_build_versions(self, other: Self) -> int:
+        if self.version is None or other.version is None:
+            return NotImplemented
```
rra, line 562: Ugh, I don't like this. It implies that the version numbers might not be comparable, but the reality is that we already know that both versions have valid semvers. We've just lost the type information by passing in too-generic of parameters. I think the calling function needs to use local variables for the semvers so that you get type narrowing when you check that they're not `None`, and then this helper should take two semvers as parameters, which avoids the need for this check entirely.

controller/src/controller/models/domain/rsptag.py
```diff
@@ -414,19 +526,50 @@ def _compare(self, other: object) -> int:
...
+                if other.version.build is not None:
+                    return -1
+                return 0
+            if other.version.build is None:
                 return 1
-            else:
-                return -1 if self.version.build < other.version.build else 1
-        else:
-            return -1 if other.version.build else 0
+        return -1 if self.version.build < other.version.build else 1
```
rra, line 572: Why did this code need to change? I didn't think we were changing anything about the build portion, but maybe I'm missing something.

controller/src/controller/models/domain/rsptag.py
athornton: somewhat rewritten given changes to rsp_[build_]version

controller/src/controller/models/domain/rsptag.py
athornton: Fixed.

athornton:
cf https://github.com/lsst-sqre/sciplat-lab/pull/205 for how we will use this in the build process.

PR #553, opened 2025-07-25
Move purger requirements to pyproject.toml

athornton:


jonathansick (changes requested):
My understanding is that purger is pip installed, so declaring dependencies in pyproject.toml without using a requirements.txt-type file is correct. However some changes in the PR contradict this fix, so it'd be good to clean those up first.

PR #565, opened 2025-08-12
Add fsadmin support

athornton:
Add a route under `/fsadmin` for the Nublado controller, to create/destroy/check-status-of a namespace and pod that has administrative access to nublado-defined volumes.

This will supersede lsst-sqre/idf-fsadmin.


controller/src/controller/services/builder/fsadmin.py
```diff
@@ -0,0 +1,149 @@
...
+        )
+
+    def _build_metadata(self, name: str, suffix: str = "") -> V1ObjectMeta:
+        """Construct the metadata for an object.
+
+        This adds some standard labels and annotations providing Nublado
+        metadata and telling Argo CD how to handle this object.
+        """
+        labels = {
+            "nublado.lsst.io/category": "fsadmin",
```
rra, line 70: Likewise here with the trailing comma.

controller/src/controller/services/builder/fsadmin.py
```diff
@@ -0,0 +1,149 @@
...
+        wanted_volumes = {m.volume_name for m in self._config.volume_mounts}
+        volumes = self._volume_builder.build_volumes(
+            (v for v in self._volumes if v.name in wanted_volumes),
+            pvc_prefix=self._config.pod_name,
+        )
+        mounts = self._volume_builder.build_mounts(
+            self._config.volume_mounts,
+            prefix=self._config.mount_prefix
+            if self._config.mount_prefix
+            else "",
```
rra, line 90: A trinary if statement here is pretty unreadable when it has to be broken across three lines. I would pull this out into a temporary variable to avoid the weird formatting.

controller/src/controller/services/_state.py
rra, line 1: Looks like you forgot to unwind this refactoring of the file server code when you simplified the fsadmin code to no longer need it.

controller/src/controller/services/fsadmin.py
```diff
@@ -0,0 +1,152 @@
...
+
+# This all looks extremely similar to the fileserver service, and
+# that's no accident.  The fsadmin environment is essentially a
+# fileserver with all the guardrails taken off, running in its own
+# namespace (but for obvious reasons, lacking a WebDAV interface).
+#
+# There's probably room for some consolidation and inheritance here,
+# but fsadmin lacks the concept of a username, so we'd have to be
+# careful about method signatures, since most of the fileserver
+# methods want a username as a parameter.
```
rra, line 11: I think this comment is now obsolete.

controller/src/controller/storage/kubernetes/fsadmin.py
```diff
@@ -0,0 +1,152 @@
...
+        ns = self._metadata.namespace
+        for pvc in objects.pvcs:
+            pvcobj = await self._pvc.read(pvc.metadata.name, ns, timeout)
+            if pvcobj is None:
+                await self._pvc.create(ns, pvc, timeout)
+
+        pod = objects.pod
+        existing_pod = await self._pod.read(pod.metadata.name, ns, timeout)
+        if existing_pod is None:
+            await self._pod.create(ns, pod, timeout)
```
rra, line 93: The pod should be recreated if it already exists, not reused. I think the same thing is true of the PVCs as well. If you get here, the pod is not working correctly, so you you have to clean it up and replace it with newly created objects or the controller will be unable to self-heal the fsadmin instance.

controller/src/controller/storage/kubernetes/fsadmin.py
```diff
@@ -0,0 +1,152 @@
...
+        pvcs = objects.pvcs
+        existing_pod = await self._pod.read(pod.metadata.name, ns, timeout)
+        if existing_pod is not None:
+            await self._pod.delete(
+                existing_pod.metadata.name, ns, timeout, wait=True
+            )
+        for pvc in pvcs:
+            if await self._pvc.read(pvc.metadata.name, ns, timeout) is None:
+                continue
+            await self._pvc.delete(pvc.metadata.name, ns, timeout)
```
rra, line 131: Just delete Kubernetes resources, don't check if they exist first. This is all handled by `KubernetesObjectDeleter` already by catching the 404 error from Kubernetes.

controller/src/controller/storage/kubernetes/fsadmin.py
```diff
@@ -0,0 +1,152 @@
...
+        We check whether the pod exists, and if so, whether it is running.
+        We do not need to check the PVCs, because if they're necessary,
+        the pod can't be running unless it has them.
+        """
+        ns = self._metadata.namespace
+        existing_pod = await self._pod.read(self._config.pod_name, ns, timeout)
+        if existing_pod is None:  # No pod
+            return False
+        # Return its running state
+        return existing_pod.status.phase == PodPhase.RUNNING
```
rra, line 152: The comment here is wrong -- this doesn't return the running state.

controller/src/controller/background.py
rra, line 1: I think these changes are stray and left over from the previous version.

controller/src/controller/config.py
```diff
@@ -498,6 +499,123 @@ class EnabledFileserverConfig(FileserverConfig):
...
+        ContainerImage,
+        Field(
+            title="fsadmin Docker image",
+            description="Docker image to run as fsadmin",
+        ),
+    ] = ContainerImage(
+        repository="ghcr.io/lsst-sqre/nublado-fsadmin",
+        pull_policy=PullPolicy.ALWAYS,
+        tag="latest",
+    )
```
rra, line 546: I don't think we should ever use floating versions like this. We already have to update so many versions in Phalanx when we do a new release that it needs to be scripted, so I don't think adding one more will matter.

controller/src/controller/config.py
```diff
@@ -498,6 +499,123 @@ class EnabledFileserverConfig(FileserverConfig):
...
+        ),
+    ] = {}
+
+    pod_name: Annotated[
+        str,
+        Field(
+            title="fsadmin pod name",
+            description="Pod/container name for fsadmin",
+        ),
+    ] = "fsadmin"
```
rra, line 577: `nublado-fsadmin` maybe?

controller/src/controller/config.py
```diff
@@ -498,6 +499,123 @@ class EnabledFileserverConfig(FileserverConfig):
...
+    volume_mounts: Annotated[
+        list[VolumeMountConfig],
+        Field(
+            title="Volume mounts",
+            description=(
+                "Volumes mounted in the fsadmin pod. The ``containerPath``"
+                " settings represent the path visible inside the container."
+            ),
+        ),
+    ] = []
```
rra, line 616: Shouldn't this just be copied from labs? Why would we want to mount the volumes in different places in the fsadmin container than in the user labs? Or are there other settings that would need to be different for the admin case? I don't really like having to duplicate this (and make sure it refers correctly to the lab volumes).

If you do keep this, it needs a validator to make sure that it's consistent with `lab.volumes`.

controller/src/controller/storage/kubernetes/fsadmin.py
```diff
@@ -0,0 +1,152 @@
...
+    directly. Even though there are not many objects, having a wrapper layer
+    might be easier to follow.
+    """
+
+    def __init__(
+        self,
+        config: FSAdminConfig,
+        metadata_storage: MetadataStorage,
+        api_client: ApiClient,
+        logger: BoundLogger,
```
rra, line 49: Nit: Over three parameters, so this should force keyword arguments with `*,` before the arguments.

controller/tests/handlers/fsadmin_test.py
```diff
@@ -0,0 +1,103 @@
...
+from kubernetes_asyncio.client import V1Pod
+from safir.testing.kubernetes import MockKubernetesApi
+
+from controller.models.domain.gafaelfawr import GafaelfawrUser
+
+from ..support.config import configure
+
+
+@pytest.mark.asyncio
+async def test_create_delete(
```
rra, line 16: Needs locking tests where you try to delete it at the same time that you're creating it. You can do that by having the pod start in pending and then start a delete and only afterwards change the pod state to running. I think there's a file server test like that.

controller/tests/handlers/fsadmin_test.py
```diff
@@ -0,0 +1,103 @@
...
+    assert r.status_code == 204
+
+    # Verify it's running
+    r = await client.get("/nublado/fsadmin/v1/service")
+    assert r.status_code == 204
+
+    # Request it again; should detect it exists and
+    # return immediately without actually doing anything.
+    r = await client.get("/nublado/fsadmin/v1/service")
+    assert r.status_code == 204
```
rra, line 42: Either the comment or the code is wrong, since this doesn't request a new pod.

controller/tests/handlers/fsadmin_test.py
```diff
@@ -0,0 +1,103 @@
...
+
+    # Check that pod was created correctly
+    namespace = "nublado"
+
+    # Check that it has a pod.
+    pods = await mock_kubernetes.list_namespaced_pod(namespace)
+    pod: V1Pod | None = None
+    for ext_pod in pods.items:
+        if ext_pod.metadata.name == config.fsadmin.pod_name:
+            pod = ext_pod
```
rra, line 96:
```suggestion
    pod = next(
        p
        for p in await mock_kubernetes.list_namespaced_pod(namespace)
        if p.metadata.name == config.fsadmin.pod_name
    )
```

docs/about/index.rst
```diff
@@ -26,6 +26,10 @@ The controller is also responsible for constructing the spawner form for the Jup
 To supplement the JupyterLab interface, which only allows simple file upload and download into the user's home file space, the Nublado controller can optionally also create WebDAV file servers for the user that mount the same file systems that user lab pods mount.
 This allows users to use the WebDAV clients built into most operating systems to more readily copy files to and from the file space underlying their lab.
 
+There is additionally a mechanism to allow creation and destruction of an administrative file server.
```
rra, line 29: "starting and stopping" maybe?

docs/admin/fsadmin.rst
```diff
@@ -0,0 +1,38 @@
+#######################
+The fsadmin environment
```
rra, line 2: "File system administration pod" maybe instead?

fsadmin/scripts/install-packages
```diff
@@ -0,0 +1,40 @@
...
+    git \
+    rsync \
+    emacs-nox \
+    vim \
+    psmisc \
+    sudo \
+    quota
+
+# Make life better. Come on, it's 2025.
+chsh -s /bin/bash root
```
rra, line 36: Nit: This isn't installing packages. Maybe move it to a `RUN` command directly in the Dockerfile?

fsadmin/screenrc
```diff
@@ -0,0 +1,4 @@
+defscrollback 10000
+backtick 0 30 30 sh -c 'screen -ls | grep --color=no -o "$PPID[^[:space:]]*"'
+hardstatus string "%{= KW} %H [%`] %{= Kw}|%{-} %-Lw%{= bW}%n%f %t%{-}%+Lw"
+hardstatus alwayslastline
```
rra, line 4: This is completely incomprehensible. Could you add some comments explaining what this does?

athornton: Well, "I stole it from The Internet 20 Years Ago" but ... OK, I've taken a crack at it.

This is why there's tmux in there too, for the fortunate people who don't have the screen commands burned into their muscle memory and can shift to something that has a configuration language that is not right-up-there-with-sendmail.cf in terms of unreadability.

athornton:
I added the extraVolumes/extraVolumeMounts because the user labs won't ever mount, for instance, the firefly home dir (except in dev it's not on /home anymore) but we might want to from fsadmin.

controller/src/controller/config.py
athornton: We already know it's in the nublado namespace.  Saves eight characters of typing when you want to exec into the pod.

controller/src/controller/storage/kubernetes/fsadmin.py
```diff
@@ -0,0 +1,193 @@
...
+            await self._pvc.create(ns, pvc, timeout)
+        await self._pod.create(ns, objects.pod, timeout)
+
+        # Wait for the pod to start.
+        await self._pod.wait_for_phase(
+            objects.pod.metadata.name,
+            ns,
+            until_not={PodPhase.UNKNOWN, PodPhase.PENDING},
+            timeout=timeout,
+        )
```
rra, line 108: It would be useful if `wait_for_phase` returned the `V1PodStatus` because then you'd have everything you needed already. Ah well, it's a trade-off for how it's used elsewhere. Since you have the status, though, you may as well put that into the `FSAdminStatus` as well.

controller/src/controller/storage/kubernetes/fsadmin.py
```diff
@@ -0,0 +1,193 @@
...
+
+        Notes
+        -----
+        If the pod exists but is not yet running, or is in the process of
+        terminating, we report None. The start time is only valid for pods
+        which are ready to accept work.
+        """
+        if not await self.is_fsadmin_ready(timeout):
+            return None
+        return self._start_time
```
rra, line 193: You can replace both of these methods with one method that returns an `FSAdminStatus` if the pod is running or `None` if it's not.

controller/src/controller/config.py
```diff
@@ -498,6 +499,135 @@ class EnabledFileserverConfig(FileserverConfig):
...
+        ),
+    ] = None
+
+    application: Annotated[
+        str | None,
+        Field(
+            title="Argo CD application",
+            description="Argo CD application into which to put fsadmin pods",
+        ),
+    ] = None
```
rra, line 521: This should always be in the same application as the controller, not independently configurable, since it's using the controller's namespace.

controller/src/controller/config.py
```diff
@@ -498,6 +499,135 @@ class EnabledFileserverConfig(FileserverConfig):
...
+        ContainerImage,
+        Field(
+            title="fsadmin Docker image",
+            description="Docker image to run as fsadmin",
+        ),
+    ] = ContainerImage(
+        repository="ghcr.io/lsst-sqre/nublado-fsadmin",
+        pull_policy=PullPolicy.IF_NOT_PRESENT,
+        tag="8.13.0",
+    )
```
rra, line 569: Defaults should be in `values.yaml` since that's where we update the version number. This should just be a required configuration parameter with no default.

controller/src/controller/services/fsadmin.py
```diff
@@ -0,0 +1,151 @@
...
+        """
+        timeout = Timeout("Filesystem admin query", self._config.timeout)
+        try:
+            async with self._lock:
+                return await self._storage.get_start_time(timeout)
+        except Exception as e:
+            msg = "Error querying fsadmin status"
+            self._logger.exception(msg)
+            await self._maybe_post_slack_exception(e)
+            raise
```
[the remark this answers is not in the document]
rra: Oh, except here you need to catch the object not found error and turn that into an exception that represents an HTTP status code of 404.

controller/src/controller/config.py
athornton: When we do that, anyone who naively upgrades the Nublado controller, without changing their values.yaml in Phalanx, breaks their system.

I can do that (and make this a major version increase) but it was an intentional choice to default these so that people's extant configs would keep working.

controller/src/controller/services/builder/fsadmin.py
```diff
@@ -0,0 +1,168 @@
...
+        volumes.extend(self._config.extra_volumes)
+        volume_mounts = [
+            VolumeMountConfig(
+                container_path=x.container_path,
+                sub_path=x.sub_path,
+                read_only=False,
+                volume_name=x.volume_name,
+            )
+            for x in self._volume_mounts
+        ]
```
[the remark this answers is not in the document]
athornton: Maybe we currently don't.  I think we did at the summit for a while?

controller/src/controller/handlers/fsadmin.py
```diff
@@ -0,0 +1,92 @@
...
+        # propagate as an uncaught exception.
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail=[
+                {
+                    "msg": f"Failed to delete fsadmin: {e!s}",
+                    "type": "fsadmin_delete_failed",
+                }
+            ],
+        ) from e
```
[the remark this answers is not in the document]
athornton: Huh.  There I was following what `delete_lab` does.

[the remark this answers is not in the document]
athornton: The other lab handlers seem to just let Kubernetes and Timeout errors propagate.

controller/src/controller/storage/kubernetes/fsadmin.py
athornton: I think it's better to return an FSAdminStatus if it's running, and throw a ClientRequestError if it isn't explaining what the problem is.  Then I can catch that in the handler and turn it into an error return.  Otherwise I have to propagate what's really just the exception information (which comes down to, "there's no pod" or "the pod's not in Running state") through some additional method, which seems dumb.

controller/src/controller/handlers/fsadmin.py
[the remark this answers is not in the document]
athornton: Alternatively, if you'd like me to remove the handler from delete_lab, I can do that instead.

rra (approved):
Looks mostly good, just a few more smaller things. Thanks for all the fixes!

controller/src/controller/storage/kubernetes/fsadmin.py
```diff
@@ -0,0 +1,190 @@
...
+        TimeoutError
+            Raised if fsadmin objects are not deleted within provided timeout.
+        """
+        ns = self._metadata.namespace
+        pod = objects.pod
+        pvcs = objects.pvcs
+        self._start_time = None
+        await self._pod.delete(pod.metadata.name, ns, timeout, wait=True)
+        for pvc in pvcs:
+            await self._pvc.delete(pvc.metadata.name, ns, timeout, wait=True)
```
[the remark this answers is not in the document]
athornton: I'm concerned about what happens if someone tries to spawn while the PVCs are being deleted.  I think there's room for a race since we don't check for the PVCs in the readiness check, so I ... well, I don't know what happens if you try to issue the new PVC creation as replacement while the PVCs are being deleted.

controller/src/controller/handlers/fsadmin.py
```diff
@@ -0,0 +1,82 @@
...
+    summary="Create fsadmin instance",
+    tags=["admin"],
+    status_code=200,
+)
+async def create_fsadmin(
+    *,
+    cmd: FSAdminCommand,
+    context: Annotated[RequestContext, Depends(context_dependency)],
+) -> FSAdminStatus:
+    _ = cmd
```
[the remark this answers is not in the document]
rra: No, there's no such thing as unused parameter warnings in Python so far as I know.

PR #572, opened 2025-08-22
tickets/DM-52281: Add apply_user_options() to spawner

athornton:
The JupyterHub spawner base class logs a warning if you have a user_options dict and don't have an apply_user_options() method to handle it.  Effectively, the Spawner base freaks out a little that it's getting unhandled options.

So this just checks that you don't have any fields in user_options that are not present in the spawner template, and errors if you do.  That should quell the warnings and give us a tiny bit of validation on top.


athornton:
Marking as draft until I get an image build, run it, and check that it doesn't break anything in dev.

PR #576, opened 2025-09-04
DM-52391: Add Sentry error reporting to controller

fajpunk:
Add Sentry error reporting and enhance controller exceptions with Sentry metadata.

Goes with these Safir PRs:
* https://github.com/lsst-sqre/safir/pull/443
* https://github.com/lsst-sqre/safir/pull/448

And this Phalanx PR:
* https://github.com/lsst-sqre/phalanx/pull/5331


fajpunk:
> Is there any way to test whether exceptions are annotated properly? I'm not sure how important that is or if it's worth it.

There are two ways:
* Manually construct an error object and assert things about the thing returned from calling `to_sentry()`, which is what I've added here
* Asserting payloads that would get sent to Sentry during integration tests of the whole app, as described in the [Safir Sentry docs](https://safir.lsst.io/user-guide/sentry.html#testing)

PR #580, opened 2025-09-10
Fix matrix build

fajpunk:


athornton (approved):
Whoops.  That was a dumb bug for me to put in there.

PR #587, opened 2025-09-26
DM-52657: Rewrite build system to use uv.lock files

rra:
Reorganize the build system into multiple workspaces so that all parts of the build can use `uv.lock` files and nox-uv for simpler `noxfile.py` files.

Create a (for now, dummy) top-level nublado package to serve as the workspace root, and include inithome and purger in that namespace since there is no pressing need for them to have separate virtual environments or frozen dependencies. Maintain the controller and the Nublado client as separate projects with their own `uv.lock` files since the controller has heavier dependencies and the client may need to support a wider range of Python versions.

Turn the hub directory into a (currently dummy) Python package named nublado-jupyterhub so that it can manage its own dependencies. Make that a workspace root and move authenticator and spawner into a plugins directory inside that workspace.

Merge the test suites of inithome and purger, and of authenticator and spawner, to follow the rule of one test suite per workspace and dev dependency management at the root of the workspace.

Use recursive nox to build all of these projects together. This allows use of nox-uv at each level and far simpler `noxfile.py` files. It would have been possible to use non-recursive nox, but the `noxfile.py` file would have been much more complex.

Use uv sync to install dependencies, except for the JupyterHub image. In that case, export the `uv.lock` file to a `requirements.txt` so that the packages can still be installed in the system paths and override the packages that come with the Zero to JupyterHub container. For the controller, inithome, and purger images, use our current virtual environment management approach for service builds.


rra:
The lint, test-client (3.13), and test-hub (3.13) checks don't exist any more after this change (they're all part of test), which is why they show as pending. I'll remove them from the merge requirements before merging this, after it's approved.

Dockerfile.purger
```diff
@@ -25,36 +25,33 @@ RUN ./install-base-packages.sh && rm ./install-base-packages.sh
 
 FROM base-image AS install-image
 
+# Install uv.
+COPY --from=ghcr.io/astral-sh/uv:0.7.15 /uv /bin/uv
```
athornton, line 29: How about we go to a current version of UV while we're up in here?

Makefile
```diff
@@ -0,0 +1,22 @@
+.PHONY: help
```
athornton, line 1: Seems weird that we didn't have a Makefile before.

tests/purger/util.py
```diff
@@ -4,9 +4,10 @@
 import os
 from pathlib import Path
 
-from rubin.nublado.purger.models.plan import FileReason
 from safir.pydantic._types import _validate_human_timedelta
 
+from rubin.nublado.purger.models.plan import FileReason
+
```
athornton, line 11: Did the import sorter change its mind, or did I manage somehow never to run it over these tests?

Makefile
rra: For a while we tried to do without one for packages using nox, but it's really useful for bootstrapping and updating dependencies.

client/src/rubin/nublado/client/nubladoclient.py
```diff
@@ -111,7 +111,7 @@ def __init__(self, event_source: EventSource, logger: BoundLogger) -> None:
         self._logger = logger
         self._start = datetime.now(tz=UTC)
 
-    async def __aiter__(self) -> AsyncGenerator[SpawnProgressMessage, None]:
```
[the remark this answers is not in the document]
rra: (This change was automatically made by the latest Ruff.) `Generator` and `AsyncGenerator` take one, two, or three arguments. The first argument is the yield type, the second is the send type, and the third is the return type. So it's also correct to add one or two `None` values in the type of a simple generator that only yields values. However, the send and return types default to `None`, so it's also unnecessary to add those, and the latest Ruff automatically removes them.

hub/pyproject.toml
```diff
@@ -20,42 +20,62 @@ classifiers = [
 ]
 requires-python = ">=3.12"
 dependencies = [
-    "httpx<1",
-    "httpx-sse<1",
-    "jupyterhub<6",
-    "PyYAML<7",
-    "traitlets<6",
+    "jupyterhub==5.3.0",
```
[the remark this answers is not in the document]
rra: (This isn't a change; it was already pinned in the hub `requirements/main.in` that this file was based on.) Updating this pin is documented in the process for going to a new JupyterHub release in the development documentation, so hopefully people will read that?

PR #589, opened 2025-09-29
DM-52657: Update pinned uv versions during make update

rra:
Copy over the script that updates pinned uv versions and update older versions in various Dockerfiles.


athornton (approved):
LGTM

PR #595, opened 2025-10-01
Rebuild jupyterlab-base with Trixie

athornton:


athornton:
Nope.  Export via TeX is much smaller.

PR #597, opened 2025-10-01
DM-52657: Fix Docker build for controller

rra:
The Docker build for the Nublado controller had some confusion between top-level and subdirectory virtualenvs and was using the wrong one for the running container. Consistently use the virtual environment for the controller and copy only it over to the final container.


athornton (approved):
Good catch.

PR #599, opened 2025-10-01
Use Trixie for jupyterlab-base, TeX for PDF export

athornton:


rra (approved):
Looks great. Thanks for working through this!

PR #601, opened 2025-10-07
DM-52784: Use Repertoire for Gafaelfawr JupyterHub plugin

rra:
Use Repertoire's service discovery library to find the URL for the Gafaelfawr logout page in the Gafaelfawr JupyterHub plugin.


athornton (approved):
After the update-deps stuff there's not much there there, and what there is looks correct.

PR #603, opened 2025-10-07
purger - Fix error when logging failures

fajpunk:
Dict keys in structlog context items  must be must be str, int, float, bool or None, otherwise we get a type error:

```
TypeError: keys must be str, int, float, bool or None, not PosixPath
```

### DIFF
--- a/purger/src/rubin/nublado/purger/purger.py
+++ b/purger/src/rubin/nublado/purger/purger.py
@@ -7,11 +7,10 @@ import asyncio
 import datetime
 from pathlib import Path
 
-import structlog
 import yaml
-from safir.logging import Profile, configure_logging
-from safir.slack.blockkit import SlackTextBlock
+from safir.logging import configure_logging
 from safir.slack.webhook import SlackRouteErrorHandler
+from structlog.stdlib import BoundLogger, get_logger
 
 from .config import Config
 from .constants import ROOT_LOGGER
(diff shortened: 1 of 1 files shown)


purger/src/rubin/nublado/purger/purger.py
```diff
@@ -7,11 +7,10 @@
 import datetime
 from pathlib import Path
 
-import structlog
 import yaml
-from safir.logging import Profile, configure_logging
-from safir.slack.blockkit import SlackTextBlock
+from safir.logging import configure_logging
 from safir.slack.webhook import SlackRouteErrorHandler
+from structlog.stdlib import BoundLogger, get_logger
```
fajpunk, line 13: Importing these tings from `structlog.stdlib` instead of `structlog` gives us type-safe versions of them... though not type-safe enough to catch this original bug ☹️

purger/src/rubin/nublado/purger/purger.py
```diff
@@ -226,18 +225,7 @@ async def _perform_report(self) -> None:
         if self._plan is None:
             raise PlanNotReadyError("Cannot report: plan not ready")
         rpt_text = str(self._plan)
-        if self._config.alert_hook is not None:
-            rpt_msg = SlackTextBlock(
-                heading="Purge plan",
-                text=rpt_text,  # May be truncated
-            )
```
fajpunk, line 233: Like we talked about:

* We don't want to send the reports to Slack
* The Slack notifucation machinery is not wired up yet anyway. It will be in a future PR.

PR #667, opened 2025-11-07
Add jupyter-server-documents

athornton:


athornton:
Making this a draft because jupyter-server-documents behaves very weirdly with symlinks and I need to do some more experimentation before we commit to it.

athornton:
superseded

PR #729, opened 2025-12-17
tickets/DM-53638: init containers

athornton:
Rework init containers for simplified configuration.

Draft for now, actual reviewable/mergeable PRs, forming a queue which eventually gets to this state, will be trickling out.  This is only a PR in order to generate containers I can use in testing.


athornton:
All of that work has been merged, now just rebuilding to get newer lsst-rsp that migrates startup into here.

athornton:
superseded

PR #733, opened 2025-12-22
tickets/DM-53703: jupyter-server-documents

athornton:
Branch for experimentation with jupyter-server-documents.  It makes the file root `/` so we can meaningfully have the same file open between different users.

It seemed earlier to do weird things when confronted with symlinks.  This branch is for making an experimental build to run some controlled experiments.

It depends on the DM-53683 stack, and should not be merged until we determine what the behavior is.


athornton:
superseded

PR #820, opened 2026-04-08
t/DM-53832-extensions

athornton:
Add newer rsp-jupyter-extensions to jupyterlab-base, freshen dependencies.


athornton:
superseded

PR #837, opened 2026-04-13
tickets/DM-53832-settings: Add explicit settings

athornton:
* enable CST landing page
* enable tutorials
* enable query menu
* choose between `/` and `$HOME` for Lab file browser root.


src/nublado/controller/services/builder/lab.py
```diff
@@ -375,7 +379,19 @@ def _build_env_config_map(
...
+        _hrfbr = "HOME_RELATIVE_TO_FILE_BROWSER_ROOT"
+        if self._config.file_browser_root == FileBrowserRoot.ROOT:
+            env[_hrfbr] = self._build_home_directory(user.username)
+        else:
+            env[_hrfbr] = "."
+        # Set up booleans previously represented by RSP_SITE_TYPE
+        if self._config.enable_rubin_query_menu:
+            env["ENABLE_RUBIN_QUERY_MENU"] = "TRUE"
+        if self._config.enable_tutorials_menu:
+            env["ENABLE_TUTORIALS_MENU"] = "TRUE"
```
rra, line 394: I really do not want to add any new environment variables to Nublado. I definitely don't want to add any environment variables that don't start with `NUBLADO_`. If this is just to remove `RSP_SITE_TYPE`, I think the prerequisite for removing it is to finish the `config.json` work.

src/nublado/controller/config.py
```diff
@@ -934,6 +944,51 @@ class LabConfig(BaseModel):
         ),
     ] = {}
 
+    enable_cst_landing_page: Annotated[
```
rra, line 947: I don't think there is anything CST-specific about the feature in Nublado. It's just a landing page that overrides the default lab landing page. Given that, I think we should use a more generic name in the Nublado code and configuration.

src/nublado/controller/config.py
```diff
@@ -934,6 +944,51 @@ class LabConfig(BaseModel):
...
+        Field(
+            title="Enable CST landing page",
+            description=(
+                "Whether or not to direct the user to the CST landing page"
+                " at Lab spawn time."
+            ),
+        ),
+    ] = False
+
+    enable_rubin_query_menu: Annotated[
```
rra, line 958: Similarly here, there's nothing Rubin-specific about this, just TAP-specific. Maybe `enable_tap_query_menu` instead?

rra (commented):
This will also need a change log and documentation updates for the new configuration settings, but per the comment it may be blocked on finishing the `config.json` work.

athornton:
Closing in favor of #845

PR #838, opened 2026-04-15
tickets/DM-54495: add migrator

athornton:
Add user-migrator to nublado.


src/nublado/controller/models/v1/migrator.py
```diff
@@ -0,0 +1,142 @@
...
+
+    new_user: Annotated[str, Field(title="Target username to copy to")]
+
+    start_time: Annotated[
+        str,
+        Field(
+            title="When the migrator pod was started",
+            examples=["2026-04-25T16:57:38.825103+00:00"],
+        ),
+    ] = datetime.datetime.now(tz=datetime.UTC).isoformat()
```
stvoutsin, line 38: Should this be a default_factory - lambda that sets the default time, otherwise all MigratorStatus instances will share the datetime that is assigned at class load time?

src/nublado/migrator/_service.py
```diff
@@ -0,0 +1,144 @@
...
+                follow_symlinks=False
+            ):
+                shutil.chown(
+                    dirpath,
+                    user=self._uid,
+                    group=self._gid,
+                    follow_symlinks=False,
+                )
+                for fn in filenames:
+                    shutil.chown(
```
stvoutsin, line 136: Should `follow_symlinks=False` should be added here or is it omitted by design? What actually gets chowned if it is a symlink and we don't tell it not to follow, will it chown the source directory file?

src/nublado/controller/models/domain/migrator.py
```diff
@@ -0,0 +1,27 @@
...
+        Username for source user to copy from.
+    new_user
+        Username for target user to copy to.
+
+    Returns
+    -------
+    str
+        Name of migrator pod.
+    """
+    return f"migrator-{old_user}-to-{new_user}"
```
stvoutsin, line 23: Do you want to sanitize the usernames here? Will dots / underscores etc cause any issues?

src/nublado/migrator/_exceptions.py
```diff
@@ -0,0 +1,61 @@
...
+        super().__init__(message)
+        self.exitcode = 5  # No target directory
+
+
+class CopyError(BaseMigratorError):
+    """Something went wrong with file copy."""
+
+    def __init__(self, message: str) -> None:
+        super().__init__(message)
+        self.exitcode = 6  # File permission error
```
stvoutsin, line 53: Looks like the comment may need to updated here to "Copy error" or whatever you prefer.

src/nublado/controller/services/migrator.py
```diff
@@ -0,0 +1,111 @@
...
+        self._storage = migrator_storage
+        self._slack = slack_client
+        self._logger = logger
+        self._lock = asyncio.Lock()
+
+    async def create(self, old_user: str, new_user: str) -> MigratorStatus:
+        """Ensure the migrator environment exists.
+
+        If we don't have a migration environment for this user pair,
+        create it.  If we do, just return. This gets called by the
```
stvoutsin, line 62: The comment that if we do have a migration we just return seems to apply only if the pod is still running. Otherwise it looks like the cache entry is cleared and the migration is re-created, so perhaps worth documenting this behaviour as well.

src/nublado/controller/handlers/migrator.py
```diff
@@ -0,0 +1,65 @@
...
+            "description": "Ownership change failed after copy",
+            "model": ErrorModel,
+        },
+        404: {"description": "User not found", "model": ErrorModel},
+        406: {"description": "File copy failed", "model": ErrorModel},
+    },
+    description=("Get the status of a user migration."),
+    summary="Get migrator status",
+    tags=["migrator"],
+    status_code=200,
```
stvoutsin, line 31: This can also return a 204 so worth updating for completeness

tests/controller/handlers/migrator_test.py
```diff
@@ -0,0 +1,202 @@
+"""Tests for migrator route."""
```
stvoutsin, line 1: Do you want to also add a couple tests that simulate a pod exiting with error (error codes 4, 5, 6 and 7)?

docs/admin/migrator.rst
```diff
@@ -0,0 +1,59 @@
...
+  The HTTP body for a ``200`` response is JSON with four fields:
+
+  #. ``start_time``, whose value is a textual representation of a UTC ISO 8601 datestamp showing the time the pod was created.
+  #. ``end_time``, whose value is a textual representation of a UTC ISO 8601 datestamp showing the time the pod exited (or ``null`` if it has not exited).
+  #. ``running``, a boolean indicating whether the pod is running.
+  #. ``exit_code``, whose value is the exit code of the migrator pod process (or ``null`` if the pod has not exited, in which case ``running`` should be ``true`` and ``end_time`` should be ``null``).
+
+  The ``GET`` returns an error code if the migration was attempted but failed for some reason:
+
+  It returns ``404`` if either the source or target user cannot be found.
```
stvoutsin, line 38: It looks like this documentation might be referring to an earlier iteration? Since you know return a 200 with the error codes in the body for errors (which I agree with by the way).

src/nublado/controller/services/builder/migrator.py
```diff
@@ -0,0 +1,183 @@
...
+        """Construct the pod for fsadmin."""
+        pod_name = build_migrator_pod_name(old_user, new_user)
+        # Volumes and volume_mounts should each be a one-item list
+        volumes = [
+            x
+            for x in self._lab_config.volumes
+            if x.name == self._lab_config.home_volume_name
+        ]
+        volume_mounts = [
+            x
```
stvoutsin, line 91: You probably want `copy(x)` here otherwise you modify the original shared volume mounts?

docs/admin/migrator.rst
```diff
@@ -0,0 +1,59 @@
...
+
+  It returns ``406`` if the file copy fails.
+
+  It returns ``403`` if there is an error changing ownership of the copied files after the copy completes.
+
+  It returns ``409`` if a migration for the same user pair, but in the opposite direction, is currently in progess.
+
+  In any case that does not return a status with ``running`` set, indicating an operation in progress, the completed pod, if any, will be removed.
+  Thus the querier gets only one chance to read the pod's exit status, which is encoded in the ``exit_code`` field of the JSON response (or, perhaps more conveniently, determined from the HTTP status code returned).
```
stvoutsin, line 47: I can see how this helps with handling cleanup and simplifying the life for the client, but it feels odd in that it is going against HTTP semantics since GET is not idempotent here.

A client that retries on a transient network failure would get a 204 on the second call and lose the result potentially.

I think we were trying be more strict in following HTTP semantics, would be to make GET a pure read with no side effects (Checks k8s, falls-back to cache but never deletes anything), thus making it idempotent and free to be retried.

Then adding a DELETE handler that the client calls after the migration is done. Perhaps you could also add a `ttlSecondsAfterFinished` to the pod as safety net after it has been completed. 

This would add one extra step to the client (calling DELETE) but since we know that we control the client as well, I think that would be fine? 

Perhaps worth discussing with whoever is writing the client for this (Jonathan?) or bringing up as a stand-up topic. It may be the case that we don't care about being HTTP strict in favor of less complexity, but my initial hunch is that the complexity isn't that high so it's worth doing?

src/nublado/controller/handlers/migrator.py
```diff
@@ -0,0 +1,65 @@
...
+        403: {
+            "description": "Ownership change failed after copy",
+            "model": ErrorModel,
+        },
+        404: {"description": "User not found", "model": ErrorModel},
+        406: {"description": "File copy failed", "model": ErrorModel},
+    },
+    description=("Get the status of a user migration."),
+    summary="Get migrator status",
+    tags=["migrator"],
```
stvoutsin, line 30: Are the tags between GET and POST here different by design, or did you want them appearing under the same tag perhaps?

stvoutsin (commented):
Overall the PR looks great, just a few comments/questions and then a couple of what seem to be bugs unless they are intended and I've misinterpreted which is very possible.

The one higher-level discussion point is on GET (+ delete), perhaps worth discussing during standup in case I'm over-complicating things for you and the client with my suggestion.

src/nublado/controller/models/v1/migrator.py
athornton: Oh.  Uh, probably?  Setting the start time as the time the *instance* is created is the intended behavior.

src/nublado/migrator/_service.py
```diff
@@ -0,0 +1,144 @@
...
+            if not self._dest_p.exists() or not self._dest_p.is_dir():
+                raise NoTargetUserDirectoryError(self._new)
+            _stat = self._dest_p.stat()
+            self._uid = _stat.st_uid
+            self._gid = _stat.st_gid
+        except BaseMigratorError as exc:
+            errorexit(exc)
+        self._logger.debug(f"Resolved paths for {self._old} and {self._new}")
+
+    def _copy_files(self) -> None:
```
rra, line 103: Would it be possible to use rsync rather than reimplementing it in Python? It's extremely difficult to write this sort of code securely (for example, Stelios's point below), and using rsync would let you use the `--chown` flag to handle ownership in a rather more straightforward way. It would probably also be considerably faster.

src/nublado/migrator/_service.py
athornton: Huh.  Good question.  The documentation is lacking.

According to the source, I probably should set it to `False`; it will try to chown the source file otherwise.  In my defense, chown() only got the parameter in Python 3.13.

src/nublado/controller/services/migrator.py
```diff
@@ -0,0 +1,111 @@
...
+        migrator_storage: MigratorStorage,
+        slack_client: SlackWebhookClient | None,
+        logger: BoundLogger,
+    ) -> None:
+        self._lab_config = config.lab
+        self._fsadmin_config = config.fsadmin
+        self._timeout = self._fsadmin_config.timeout
+        self._builder = MigratorBuilder(
+            config=config, lab_builder=lab_builder, logger=logger
+        )
```
rra, line 52: We're using dependency injection for everything else, we may as well use it for this too for consistency. (In other words, the factory should create the `MigratorBuilder` and pass it in as an argument.)

src/nublado/controller/models/domain/migrator.py
athornton: Mmmph.  I guess I need to look at USERNAME_REGEX and compare to what legal pod names are.  Maybe.

src/nublado/controller/services/migrator.py
athornton: Yeah.  The intended action is that the migration is re-run if the pod is no longer running.

docs/admin/migrator.rst
athornton: I don't think I do.  `raise_for_status()` turns that object into a 4xx error.

tests/controller/handlers/migrator_test.py
athornton: I added those in the model test, didn't I?

src/nublado/controller/models/v1/migrator.py
```diff
@@ -0,0 +1,142 @@
...
+            )
+        if self.end_time is not None:
+            et = datetime.datetime.fromisoformat(self.end_time)
+            st = datetime.datetime.fromisoformat(self.start_time)
+            elapsed = (et - st).total_seconds()
+            if elapsed < 0:
+                raise RuntimeError(
+                    f"Start time {self.start_time} is after end time"
+                    f" {self.end_time}"
+                )
```
rra, line 142: I would reserve `RuntimeError` for some sort of "this can't possibly happen" internal logic error. These are input errors because the structure of the Kubernetes resource didn't look like what we expected, so should get their own appropriate custom Nublado exception if we don't already have one for that. (That exception should also inherit from the normal `SlackException` stuff so that we get reasonable error reporting.)

src/nublado/controller/handlers/migrator.py
athornton: Not by design.  WIll fix.

src/nublado/controller/handlers/migrator.py
```diff
@@ -0,0 +1,65 @@
...
+__all__ = ["router"]
+
+
+@router.get(
+    "/migrator/v1/service",
+    responses={
+        403: {
+            "description": "Ownership change failed after copy",
+            "model": ErrorModel,
+        },
```
rra, line 24: That's not what an HTTP 403 means. The status code is reserved for an authentication error at the HTTP layer (meaning that the client that tried to send the GET does not have permission to do so). I don't think this method ever returns a 403 anyway?

src/nublado/controller/handlers/migrator.py
```diff
@@ -0,0 +1,65 @@
...
+        404: {"description": "User not found", "model": ErrorModel},
+        406: {"description": "File copy failed", "model": ErrorModel},
```
rra, line 26: Likewise, HTTP 406 is only for HTTP content negotiation. You aren't supposed to return fine-grained status in HTTP status codes; that's not really what they're for. All failures of the underlying migration operation should be returned in the JSON document with a 200 HTTP status code; the client has to look at the JSON body to see if the operation succeeded or not.

src/nublado/controller/handlers/migrator.py
```diff
@@ -0,0 +1,65 @@
...
+
+@router.post(
+    "/migrator/v1/service",
+    responses={
+        403: {
+            "description": "Ownership change failed after copy",
+            "model": ErrorModel,
+        },
+        404: {"description": "User not found", "model": ErrorModel},
+        406: {"description": "File copy failed", "model": ErrorModel},
```
rra, line 53: Same issues here with response codes.

src/nublado/controller/models/v1/migrator.py
```diff
@@ -0,0 +1,142 @@
...
+
+class MigratorStatus(BaseModel):
+    """Status for a migrator pod."""
+
+    old_user: Annotated[str, Field(title="Source username to copy from")]
+
+    new_user: Annotated[str, Field(title="Target username to copy to")]
+
+    start_time: Annotated[
+        str,
```
rra, line 33: Type here should be `datetime`.

src/nublado/controller/models/v1/migrator.py
```diff
@@ -0,0 +1,142 @@
...
+    @field_validator("start_time", "end_time", mode="before")
+    @classmethod
+    def _validate_start_time(cls, v: str | None) -> str | None:
+        if v is None:
+            return None
+        try:
+            _ = datetime.datetime.fromisoformat(v)
+        except Exception:
+            raise ValueError(f"{v} cannot be parsed as a datestamp") from None
+        return v
```
rra, line 65: This is unnecessary; Pydantic does all that automatically.

src/nublado/controller/models/v1/migrator.py
```diff
@@ -0,0 +1,142 @@
...
+    start_time: Annotated[
+        str,
+        Field(
+            title="When the migrator pod was started",
+            examples=["2026-04-25T16:57:38.825103+00:00"],
+        ),
+    ] = datetime.datetime.now(tz=datetime.UTC).isoformat()
+
+    end_time: Annotated[
+        str | None,
```
rra, line 41: Likewise here, this should be a datetime.

src/nublado/migrator/_service.py
athornton: We could....this does make us no longer portable to MacOS (uses openrsync, which doesn't have `--chown`) but for the nublado case specifically I don't think we care.  I'm not sure it's going to be faster since it's going into a timestamped directory (that is, not doing a differential copy), and presumably shutil is just doing a recursive copy.

(At least I think that's the right way to go, although you could make a case that a rerun should give you just one target directory, which is "whatever we copied at the moment you invoked the migrator" and could be updated if someone was like "but I went and used my old account for a week and now all my really important stuff is in it but I have been using my new account for a month".)

I'm not generally a fan of shelling out from my python, but I guess if the migrator basically just builds an rsync command and then `exec()s` it, that's no different than how we usually treat `git`.  There is at least one librsync wrapper but it doesn't look well-maintained or indeed very useful given that we're looking to just copy a tree.

src/nublado/cli.py
```diff
@@ -315,6 +316,14 @@ def landingpage() -> None:
         logger.exception("Provisioner failed")
 
 
+@main.command()
+def migrator() -> None:
+    """Copy a user's files from their old homespace to their new homespace."""
+    # Don't catch exceptions: we want to use the container exit code.
+    migrator = Migrator.from_env()
+    migrator.go()
```
rra, line 324: I don't really like how all the parameters are provided only by environment variables, since it makes it hard to run this manually from an fsadmin pod, for instance. I know the arguments are a little annoying since you have to provide both usernames and both home directories, but I'd rather see this as a proper command-line tool that gets its arguments on the command line. Maybe `old-user:/home/old-user new-user:/home/new-user` or something. Then the pod could be configured to just pass in command-line arguments and not muck about with environment variables.

src/nublado/migrator/_exceptions.py
```diff
@@ -0,0 +1,61 @@
...
+    "CopyError",
+    "CopyPermissionError",
+    "NoSourceUserDirectoryError",
+    "NoTargetUserDirectoryError",
+    "errorexit",
+]
+
+
+def errorexit(exc: BaseMigratorError) -> Never:
+    print(repr(exc), file=sys.stderr)
```
rra, line 17: Please don't print unstructured data to stderr in any pod. It can't be parsed by Log Explorer and then we can't easily search for it, etc. Logging should always use structlog so that it's in JSON, and in this case we'll want the old and new usernames and the old and new home directories, etc., in structured fields. (It's probably easier to do that by dumping this function and just using `logger.exception` with an appropriate bound logger with the right fields.)

src/nublado/migrator/_service.py
```diff
@@ -0,0 +1,144 @@
...
+    def from_env(cls) -> Self:
+        try:
+            _old = os.environ.get("NUBLADO_OLD_USER")
+            if not _old:
+                raise NoSourceUserDirectoryError("Unknown old user")
+            _new = os.environ.get("NUBLADO_NEW_USER")
+            if not _new:
+                raise NoTargetUserDirectoryError("Unknown new user")
+        except BaseMigratorError as exc:
+            errorexit(exc)
```
rra, line 74: Here and below, please log the exception with structlog so that we can see the backtrace if needed in Log Explorer.

src/nublado/controller/storage/kubernetes/migrator.py
```diff
@@ -0,0 +1,284 @@
...
+                old_user=old_user,
+                new_user=new_user,
+                start_time=start_time,
+                end_time=term.finished_at.isoformat(),
+                running=False,
+                exit_code=term.exit_code,
+            )
+        # Store the status in the cache.
+        async with self._lock:
+            self._cache[pod_name] = ms
```
rra, line 278: This is completely equivalent to:
```suggestion
        self._cache[pod_name] = ms
```
asyncio locks only affect code that calls something with await while the lock is held, because async code can only be pre-empted at await.

docs/admin/migrator.rst
athornton: We should ask Jonathan.

I feel like there really isn't a `DELETE` from the caller's point of view.  They're submitting a batch job, and once you kick it off, you can't stop it.  All you can do is ask for its status, and polling that status is the most straightforward way to do that, until the pod is no longer running.

I need to clarify the doc: if you ask for status, the pod is deleted if it has completed, but the status goes into the cache.  If you ask for status _after a controller restart_ when a pod had already completed _and you had already asked for status before the controller exited_ so the pod was deleted, it will no longer be in the cache and you're going to erroneously think you need to rerun the migration.  But if you hadn't asked for status already, it's going to ask K8s, and come back with a 200-still-running or it's-finished-here's-the-code.

We could make the cache persistent with Redis, I suppose, but honestly it feels like that's going to be a very rare case (once you receive the update that the pod is completed, successfully or not, you should pretty much immediately mark that in the UI).

src/nublado/controller/storage/kubernetes/migrator.py
```diff
@@ -0,0 +1,284 @@
...
+        self._logger = logger
+        self._metadata = metadata_storage
+        self._pod = PodStorage(api_client, reconnect_timeout, logger)
+        self._pvc = PersistentVolumeClaimStorage(
+            api_client, reconnect_timeout, logger
+        )
+        # Cache holds status for completed pods only, so that we can delete
+        # the pod once we have read its status.
+        self._cache: dict[str, MigratorStatus] = {}
+        self._lock = asyncio.Lock()
```
rra, line 68: This is a bit subtle, but I think the cache is misplaced here. The idea of a storage layer is to be a reusable component that provides a higher-level API to some underlying data store. In this case, that's Kubernetes; mixing in a second data store (an in-memory cache) is confusing. I would keep the cache of completed actions at the service layer.

Regardless of where this is put, if the controller restarts, we lose the history of all jobs. I suppose we can live with that, but that's going to be a bit weird if we are telling CST to come back and see if the job is finished later, since the controller may be restarted randomly at any time by Kubernetes.

src/nublado/migrator/_service.py
athornton: And I guess my blithe assumption that shutil is about as fast as the underlying C libraries it's using is not at all safe, since it is very evidently the case that curl beats the hell out of httpx (from my S3-bucket-performance testing).

src/nublado/controller/models/domain/migrator.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class MigratorObjects(FSAdminObjects):
+    """All of the Kubernetes objects making up a migrator."""
```
rra, line 27: Should the migrator be a job rather than a bare pod? It seems reasonable to restart it if it's killed by Kubernetes for some reason. (Even as a job, it might block scaledown, which is not ideal since hopefully we could restart it where it left off, particularly if it uses rsync.)

src/nublado/migrator/_service.py
rra: Yeah, I would be really surprised if shutil were anywhere as fast as rsync. rsync is making full use of optimized low-level C APIs and I'm dubious anyone bothered to implement all of that in Python.

src/nublado/controller/models/v1/migrator.py
rra: What's written there now will always use the time at which the controller starts, since that default is set during the initial import of that module. If you want to get a new time on each object creation, you do indeed need `default_factory`.

src/nublado/controller/models/domain/migrator.py
athornton: Well, restarting won't help *if* we keep the datestamped directory we're copying to (unless we pass the datestamp as one of the arguments?  ew?)...but I'm not sure we should, especially if we use rsync.  Maybe just /home/new/copied-from-old and update it if we have to?

src/nublado/controller/storage/kubernetes/migrator.py
athornton: Oh, I figured we weren't going to ask CST to come back, per se.  I figured there'd be a poll in the client service, and it'd display something like
```
[old_input] [new_input] [Submit]

old1 new1 [failed]
old2 new2 [succeeded]
old3 new3 [still running]
...
```

src/nublado/controller/storage/kubernetes/migrator.py
athornton: Should that be `with self._lock:` or just not have a lock?  I guess probably since we're using coroutines rather than threading, we won't really be contending for self._cache, right?

src/nublado/controller/storage/kubernetes/migrator.py
rra: Right, but if the controller restarts, old1 (for instance) will just not be mentioned at all, right?

src/nublado/controller/storage/kubernetes/migrator.py
rra: No lock -- you only need a lock in async code if you might call the same chunk of code from multiple tasks *and* that code uses await. If it doesn't use await, you're guaranteed that one task will execute all of the code before any other task gets to run.

docs/admin/migrator.rst
stvoutsin: I see `raise_for_status()` being called for the create path, but not for `get_status`. The GET handler returns a `MigratorStatus` unless I missread that. Otherwise if the intent is to turns those to 4xx I do think it would be better to return 200 with the error in the response body since 4xx should generally be for client errors specific to that request rather than the underlying job.

docs/admin/migrator.rst
stvoutsin: I was implying that there is a call to DELETE the pod after it is finished rather than to cancel it. It may indeed add client-side complexity that we don't want, in which case is there not a way to tell Kubernetes to delete the pod after it has completed? And then we still have knowledge if the completion status in the cache?

tests/controller/handlers/migrator_test.py
stvoutsin: Exit codes are tested in isolation, I was thinking a full handler-level test that would simulate a pod failing with an exit code. This would then verify whether we are getting a 200 or 4xx response.

PR #860, opened 2026-04-28
tickets/DM-49034: Install jupyter-ai and one or more agents in JupyterLab

athornton:
Install jupyter-ai and one or more agents in JupyterLab.


jupyterlab-base/pyproject.toml
```diff
@@ -49,6 +49,10 @@ lab = [
     "ipyvuetify",
     "ipywebrtc",
     "ipywidgets",
+    "jupyter-ai",
+    "jupyter-ai-jupyternaut",
+    "jupyter-ai-magic-commands",
+    "jupyter-ai-litellm",
```
rra, line 54: Nit: Should be sorted.

jupyterlab-base/scripts/install-base-packages
```diff
@@ -26,8 +26,11 @@ apt-get update
 # curl: CLI http client is useful
 # nano ... ed: enough editors to cover most people's habits
 #  (I gave up on emacs in the base container, alas)
+# Nodejs is needed to build JupyterLab extensions from source.
```
rra: Building from source should happen during the build step, so it should be fine to leave those dependencies in install-dependency-packages.

jupyterlab-base/scripts/install-base-packages
```diff
@@ -26,8 +26,11 @@ apt-get update
...
+# Npm is required for Gemini and Codex agents.
```
rra: It needs npm installed locally at *runtime*? That's rather weird -- it doesn't have its own frozen dependencies? What is it using npm for? (I'm asking because node + npm is going to bloat the image size.)

changelog.d/20260526_093128_athornton_DM_49034.md
```diff
@@ -0,0 +1,8 @@
+<!-- Delete the sections that don't apply -->
+
+### New features
+
+- Add Claude, Codex, Copilot, and Gemini CLI.
+
+- Add Jupyterlab support for those agents via chat.
+
```
rra, line 8: Missing a change log entry for the change to the menu wording in 0.24.1. Normally I wouldn't say that warrants a change log entry, but we have so many screenshots that we should probably note things that invalidate CST screenshots. I wouldn't call it a breaking change, though, just a other change or bug fix.

jupyterlab-base/scripts/install-base-packages
athornton: I need to update this comment.  Pretty sure node is a runtime dependency for some of the agents.

jupyterlab-base/pyproject.toml
athornton: That's weird.  I wonder why isort didn't catch that for me.

changelog.d/20260526_093128_athornton_DM_49034.md
athornton: Ah, right, that change is in the rsp-jupyter-extensions changelog, but didn't make it through to nublado.

athornton: (also, I updated that wording again in 0.25.0, to include `Autosave` because Frossie felt that otherwise it would cause undue anxiety in the user base)

jupyterlab-base/scripts/install-base-packages
athornton: Hm.  Gemini CLI is definitely written in TypeScript, so I think we need node at runtime for it and Codex.  But you're probably right, we only need npm to install.  I'll shuffle that around.

athornton: Although it looks like Codex is written in rust and the node layer just chooses which architecture binary to run.  Gemini actually depends on having a Javascript interpreter, though.

athornton: I think npm must provide some kind of mechanism like the Python entrypoint thing, but making symlinks seems to work well enough.  Not sure that saving the space matters given how huge the agents (which are usually giant statically-linked binaries) are.

athornton:
@rra I would also be sympathetic to the argument that none of this is appropriate for Jupyterlab-base and it should all go into sciplat-lab.  I put it here because a lot of it is clearly not related to the payload (the agents that end up in the system path, the Jupyter chat integrations) but ...if you think we should keep the base image slimmer and bloat up the sciplat-lab image instead, I'm not going to fight.

rra (approved):
Looks good to me!

athornton:
Alas, this somehow breaks the ability to chat with Claude and Codex.  The npm thing isn't right.

rra (approved):

PR #865, opened 2026-05-08
Require rsp-jupyter-extensions 0.23

athornton:
* Remove saving from Container Shutdown menu, because...
* ...jupyter-server-docs continually autosaves.


changelog.d/20260508_134041_athornton_DM_49034_serverdoc.md
```diff
@@ -0,0 +1,7 @@
+### Backwards-incompatible changes
+
+- Remove manual saving from Hub interaction extension.
```
rra: I think this should mention jupyterlab-base and say that this is a change to an extension that is added to JupyterLab. As written, it sounds like a change to the JupyterHub version built by Nublado. Also, "remove manual saving" I assume means "removed the Save menu item from the File menu" or something like that. Being more specific would help, since I think this is backwards-incompatible mostly because it invalidates tutorials and ideally one of the target audiences for the changelog is CST.

rra (approved):
Change itself is fine, some comments on the change log.

PR #875, opened 2026-05-29
Add 'text' param to socket.send (and mock)

athornton:


client/src/rubin/nublado/client/_session.py
```diff
@@ -164,7 +164,7 @@ async def run_python(
         # Send the message and consume messages waiting for the response.
         result = ""
         try:
-            await self._socket.send(json.dumps(request))
+            await self._socket.send(json.dumps(request), text=False)
```
rra, line 167: Let's instead do:
```suggestion
            await self._socket.send(json.dumps(request).encode())
```
This should produce the same behavior and is more explicit, I think.

client/src/rubin/nublado/client/_mock.py
```diff
@@ -923,7 +923,7 @@ def __init__(
     async def close(self) -> None:
         """Simulate close of the WebSocket."""
 
-    async def send(self, message_str: str) -> None:
+    async def send(self, message_str: str, text: bool | None = None) -> None:
```
rra, line 926: With the below change, this (and the Ruff override) are no longer needed and there are no changes to the mock, although the signature should change from `str` to `bytes` (and the argument name as well, probably), and it would probably be worth adding:
```python
        # JupyterLab, at least as of 4.5.7, only accepts Binary frames
        assert isinstance(message_str, bytes)
```
to make sure we don't regress.

athornton:
This wasn't actually the problem.  Closing.

PR #876, opened 2026-06-01
DM-55115: Update the JupyterLab WebSocket protocol

rra:
Current versions of JupyterLab appear to use a new WebSocket protocol that has a binary header with offsets in front of the JSON body. Send and receive that new protocol instead of sending and expecting raw JSON in rubin.nublado.client. Update the mock to expect and return the same.


athornton (approved):
Looks good to me, modulo, *if* recommended starts failing when we do this, then as @rra said we will need to figure out a compatibility flag.

I am reasonably hopeful that recommended is new enough that it will just quietly accept the newer protocol (although it is not clear to me yet even where the change that dropped accepting raw JSON was).  But given that the protocol document hasn't changed in three years, maybe it will just work?

athornton (approved):
Looks fine, but there are a few things I don't understand.

1) Why not just use the serialize/deserialize functions from `jupyter_server.services.kernels.connections.base` ?

2) It feels like we should, in addition to looking for str/bytes to decide whether it's raw JSON or a Jupyter Server Protocol message, be looking at the structure of the bytes to determine whether it's going to be v1 or the older binary protocol.  Are we not doing that because the str was working up until we introduced Jupyter Server Documents and it's the fact that that's recasting everything into v1 that caused the problem in the first place, so empirically we never have used and presumably never will use the older binary format?

3) is the fact that we've been handling errors wrong up to this point going to bite us?  I assume that the older implementation was based on some kind of observation of the traffic we were getting, which probably means that the older error handling matched what we were actually receiving in the wild at some point.

rra:
1. The API of those two functions is quite annoying and would require a wrapper anyway, and also I don't really want to make the Nublado client depend on the Jupyter server. In theory, those functions should never change (because if they changed, the protocol would no longer be v1), so I'm not sure we gain anything by trying to use them.
2. We are now negotiating the `v1.kernel.websocket.jupyter.org` subprotocol so we're guaranteed to never get the old binary protocol (the WebSocket connection will fail instead). Really, I should just error out if we get a Text message with JSON in it, but eh, this was only one more line of code and is probably correct.
3. I'm fairly sure we were expecting a structure from version 4 or earlier of the protocol because a bunch of stuff we were looking for was just flat-out wrong. We were looking for `msg_type` in entirely the wrong place, for instance. I'm guessing that the server sticks a few duplicate things in various places for backward compatibility, but this is now (hopefully) implementing protocol version 5 correctly according to the docs.

client/src/rubin/nublado/client/_session.py
```diff
@@ -237,24 +247,31 @@ def _parse_message(
...
+            case "execute_reply":
+                status = data["content"]["status"]
+                if status == "ok":
+                    return _JupyterOutput(content="", done=True)
+                else:
+                    traceback = data["content"].get("traceback")
+                    raise NubladoExecutionError(
+                        self._username,
+                        status=status,
+                        error="".join(traceback) if traceback else None,
```
athornton, line 254: Shouldn't that be either a space or a "\n"?  Or does each traceback item end in a newline already?

client/src/rubin/nublado/client/_websocket.py
```diff
@@ -0,0 +1,115 @@
+"""JupyterLab WebSocket protocol.
+
+See `JupyterLab WebSocket protocol`_ for more details. The
+jupyter-server-documents extension appears to only implement the
```
athornton: We can state this more strongly, since they document they only do v1: https://github.com/jupyter-ai-contrib/jupyter-server-documents/blob/0bfaafdf91651f2a888f9be49275b6d59bde0026/jupyter_server_documents/kernels/websocket_connection.py#L11

client/src/rubin/nublado/client/_websocket.py
```diff
@@ -0,0 +1,115 @@
...
+    ------
+    KeyError
+        Raised if the message is malformed.
+    ValueError
+        Raised if the message is malformed.
+    """
+    if isinstance(message, str):
+        return json.loads(message)
+
+    # Assume the message is in the v1.kernel.websocket.jupyter.org protocol
```
athornton, line 43: I guess we'll know it if we ever start getting messages in old-binary format, and can add that detection and handling if it ever happens.

athornton (approved):
Looks good.  A couple of nits, not important.  I understand why you reimplemented the deserialization function.

client/src/rubin/nublado/client/_session.py
rra: They end in newlines already so far as I can tell. The format is weird. I don't know why they did it that way. But it's been that way for a while (this code isn't new, just moved around).

client/src/rubin/nublado/client/_websocket.py
athornton: Or, like you explained, we just raise an error, and if JupyterLab ever stops supporting v1 we got work to do anyway.

PR #881, opened 2026-06-02
DM-55115: Support a timeout in run_python

rra:
Add an optional timeout parameter to the `run_python` method of `JupyterLabSession` so that users of the Nublado client can enforce a timeout on the execution of each block of code.

Add an optional `delay` parameter to the `register_python_result` of `MockJupyter` to tell the JupyterLab mock to wait before returning execution results, allowing users of the mock to test timeout handling.


client/src/rubin/nublado/client/_exceptions.py
```diff
@@ -348,6 +349,96 @@ def to_slack(self) -> SlackMessage:
...
+        *,
+        code: str | None = None,
+        context: CodeContext | None = None,
+        failed_at: datetime | None = None,
+        started_at: datetime | None = None,
+    ) -> None:
+        if started_at:
+            failed_at = failed_at or datetime.now(tz=UTC)
+            timeout = (failed_at - started_at).total_seconds()
+            msg = f"Code execution timed out after {timeout}s"
```
athornton: Dunno if it's worth doing, but `%.03d` or something might be nicer than six or nine digits of precision.

athornton (approved):
LGTM and I think this will make Mobu a lot better.

client/src/rubin/nublado/client/_exceptions.py
rra: Yeah, good idea.

PR #886, opened 2026-06-09
Remove lckr-jupyterlab-variableinspector

athornton:
It has never been compatible with JupyterLab 4.x and may be the cause of kernel hangs: https://github.com/jupyterlab-contrib/jupyterlab-variableInspector/issues/355


athornton:
Closing in favor of #887 (wrong branch)

PR #887, opened 2026-06-09
t/DM-53832-drop-variableextension: Remove lckr-jupyterlab-variableinspector

athornton:


athornton:
Starting over, since it's easier than months of rebasing.

athornton:
It's no longer present in jupyterlab-base on main, so I guess this got done.

PR #897, opened 2026-06-25
Update docs to match current reality

athornton:


docs/dev/custom-lab-behavior.rst
```diff
@@ -56,17 +56,23 @@ Where to install?
 The question of whether your changes need to be in the UI Python environment, or the DM Stack Python environment, or both, is tricky and you will need to think about it.
 If your extension is purely about controlling the Lab's behavior, and doesn't need to refer to anything inside a running notebook (which is usually the case) then it only needs to run in the UI environment.  If it presents Python functionality to the user or relies on data coming from inside the DM stack, it will need to run in the payload environment.
 
-To install a package into the UI environment, go down to the bottom of the ``install-rsp-user`` script.
-You will see the line that activates the UI virtual environment: ``source /usr/local/share/jupyterlab/venv/bin/activate``.
-Below there, do a ``uv pip install`` of your updated package or packages from the GitHub branch.
-This will look something like::
+To install a package into the UI environment, you will need to add it to pyproject.toml.
+The process is not particularly well-documented.
```
rra: Rather than adding a comment to the documentation saying something isn't documented, shouldn't we just document it? Isn't that what this PR is?

rra (approved):
Documentation changes look good, modulo the added comment that what I think this PR documents isn't documented.

docs/dev/custom-lab-behavior.rst
athornton: I think what I mean, and I'll rephrase it, is that "the recipe to install a package from a github branch via pyproject.toml is difficult to find and not spectacularly clear."

